PR #1, opened 2018-04-13
Create projectcharter.md

ju-liem:
first (very) rough draft of project charter

### DIFF
--- a/projectcharter.md
+++ b/projectcharter.md
@@ -1,0 +1,84 @@
+# forms.gov Project Charter
+Constituents of the Federal government use some number of more than 23,000 different forms available to them to interact with agencies; these agencies use the information from these forms to provide information or benefits, but spend innumberable hours recreating and quality checking handwritten entries on the original paper forms. This method of manual form processing based on paper-only forms leads to more than 11.4 billion hours of paperwork annually. As we have seen in multiple USDS engagements, web-based forms created through user-centered design methods reduce the respondent burden while producing well-structured data to the agencies for ingestion. This project creates a foundation for optimizing the process of building more useful, modern, and frictionless web-based forms.
+  
+  
+## Stakeholders
+#### Who needs to be informed about progress in major decisions?
+* Sponsor - Suzette King (OFCIO)
+* USDS Leadership
+* DHS Digital Service team
+* Others?
+
+## Project communication expectations
+**What is the best method for staying in touch?**
+* Hipchat - Paperless.gov channel
+* GitHub - us-forms-system repo
+* Email
+* Others?
+
+## What should status reports include?
+Progress towards project goals, blockers
+
+## Key Players
+* Julie Meloni | Project Lead
+* Ellen Butters | UX Designer
+* Anne Krainic | Engineer
+* Ju-Lie McReynolds | Product Manager
+* Others?
+
+## Project goal
+#### What will be the end result of the project?  
+A foundation for optimizing the process of building more useful, modern, and frictionless web-based forms through the creation of a public open-source forms library that will incorporate the U.S. Web Design Standards and include guidance and best practices for use.
+
+## Scope
+#### What is in scope?
+1. The aggregation and public sharing of lessons-learned from multiple years of USDS projects involving the transformation of forms from paper to web-based.
+2. The creation and public open-source release of a forms library based on the DSVA extensions to RSJF. This library could be used government-wide, and will incorporate the U.S. Web Design Standards along with guidance for data storage and integration best practices.
+3. The use of this library in production by at least one agency (besides VA).
+4. The creation of exemplars for conversion of high-value paper-based forms to web-based forms using this library.
+
+#### What is not in scope?
+* Support for additional form builders or rendering non-React components
+* Integration with other government digital services (e.g. login.gov)
+* Others?
+
+## Milestones
+#### …including deadlines
+* Worthy thing 1 - with date
+* Worthy thing 2 - with date
+
+## Dependencies
+#### Are there other projects/people upon whom the success of this project relies?
+* Non-team person 1
+* Other team 1
+* Other team 2
+
+## Risks to completion
+#### What are the potential risk factors that could keep us from meeting the project goal(s)?
+* Risk 1
+* Risk 2  
+
+#### What do we assume we can rely on to achieve the project goal(s)?
+* Executive support (?)
+* Team time commitments (?)
+  * Both Anne and Ju-Lie will remain 1x allocated for the full period of this project
+  * Ellen will remain available at .5x for this project
+  * Julie will continue to have continue to have .2x availability to contribute
+* Contracted resources will be available to support the engineering efforts
+
+## Exit criteria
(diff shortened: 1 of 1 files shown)


projectcharter.md
```diff
@@ -0,0 +1,84 @@
+# forms.gov Project Charter
+Constituents of the Federal government use some number of more than 23,000 different forms available to them to interact with agencies; these agencies use the information from these forms to provide information or benefits, but spend innumberable hours recreating and quality checking handwritten entries on the original paper forms. This method of manual form processing based on paper-only forms leads to more than 11.4 billion hours of paperwork annually. As we have seen in multiple USDS engagements, web-based forms created through user-centered design methods reduce the respondent burden while producing well-structured data to the agencies for ingestion. This project creates a foundation for optimizing the process of building more useful, modern, and frictionless web-based forms.
```
annekainicUSDS: I may not use the word "constituents", because more people than just voters interact with government forms. Maybe the way to phrase the first two sentences is: "The federal government maintains around 23,000 forms, which are used to carry out its essential functions and provide critical services and benefits."

The other point is that the problem isn't just paper forms; there are many "digital" forms that are actually just fillable pdfs, and still require manual processing because the data isn't collected in a format that can be automatically processed. I would amend the language here to be a little less focused on just paper forms, and more generally on manual data validating and processing, from both paper and digital pdf forms.

I might also add that the distinguishing factor of this library isn't it's ability to build web-based forms; there are many ways of doing that. One thing that is unique is the speed at which one can build them using this library. Our team went from building a form in 5 months to 5 days. This will save the government time and money in moving to a more modern framework. Just a thought!

Finally, I would edit the final sentence to be: "This project creates a tool for building complex web-based forms significantly faster than existing methods, using best practices in user experience and data collection, validation, and transmission."

projectcharter.md
```diff
@@ -0,0 +1,84 @@
...
+  
+  
+## Stakeholders
+#### Who needs to be informed about progress in major decisions?
+* Sponsor - Suzette King (OFCIO)
+* USDS Leadership
+* DHS Digital Service team
+* Others?
```
annekainicUSDS: I might add the Vets.gov team; this is where the library was originally created, and as Vets.gov is the only team currently using this tool in production, so any changes or enhancements we make to it will have to continue to work for them.

projectcharter.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Project communication expectations
+**What is the best method for staying in touch?**
```
annekainicUSDS: Do you want to perhaps specify what each communication channel is used for? I would say Hipchat is used for daily communication amongst team members, GitHub for tracking work, and Email for communication with people outside of our team (OMB folks, for example).

projectcharter.md
```diff
@@ -0,0 +1,84 @@
...
+* Email
+* Others?
+
+## What should status reports include?
+Progress towards project goals, blockers
+
+## Key Players
+* Julie Meloni | Project Lead
+* Ellen Butters | UX Designer
+* Anne Krainic | Engineer
```
annekainicUSDS: Last name is spelled "Kainic", no "r" :)

projectcharter.md
```diff
@@ -0,0 +1,84 @@
...
+## Project goal
+#### What will be the end result of the project?  
+A foundation for optimizing the process of building more useful, modern, and frictionless web-based forms through the creation of a public open-source forms library that will incorporate the U.S. Web Design Standards and include guidance and best practices for use.
+
+## Scope
+#### What is in scope?
+1. The aggregation and public sharing of lessons-learned from multiple years of USDS projects involving the transformation of forms from paper to web-based.
+2. The creation and public open-source release of a forms library based on the DSVA extensions to RSJF. This library could be used government-wide, and will incorporate the U.S. Web Design Standards along with guidance for data storage and integration best practices.
+3. The use of this library in production by at least one agency (besides VA).
+4. The creation of exemplars for conversion of high-value paper-based forms to web-based forms using this library.
```
annekainicUSDS: What does this mean?

projectcharter.md
```diff
@@ -0,0 +1,84 @@
...
+
+#### What is not in scope?
+* Support for additional form builders or rendering non-React components
+* Integration with other government digital services (e.g. login.gov)
+* Others?
```
annekainicUSDS: Do we want to list features that will not be supported in the first release? Such as saving a form in progress, prefilling existing information, back-end connections to internal services, non-linear form progress (this is one we haven't fully discussed yet).

projectcharter.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Dependencies
+#### Are there other projects/people upon whom the success of this project relies?
+* Non-team person 1
+* Other team 1
+* Other team 2
+
+## Risks to completion
+#### What are the potential risk factors that could keep us from meeting the project goal(s)?
+* Risk 1
```
annekainicUSDS: Not getting additional engineering resources

projectcharter.md
```diff
@@ -0,0 +1,84 @@
...
+* Risk 2
```
annekainicUSDS: Not solidifying an agency to partner with outside of VA (we know DHS is a potential partner, but we haven't been able to solidify that relationship yet)

projectcharter.md
```diff
@@ -0,0 +1,84 @@
...
+
+#### What do we assume we can rely on to achieve the project goal(s)?
+* Executive support (?)
+* Team time commitments (?)
+  * Both Anne and Ju-Lie will remain 1x allocated for the full period of this project
+  * Ellen will remain available at .5x for this project
+  * Julie will continue to have continue to have .2x availability to contribute
+* Contracted resources will be available to support the engineering efforts
+
+## Exit criteria
```
annekainicUSDS: This is interesting. This framework sort of assumes that there's a finality to the project, that it can end with a specific goal being achieved. However, I don't see this work ever really ending. Since we'll be creating an open source library, it will have to be maintained over time, new features will probably need to be added, bugs/issues fixed. I think our exit criteria is rather figuring out a long-term maintenance plan and transferring responsibility to those people.

projectcharter.md
ju-liem: I pulled this from what was included in priorities.md. @jcmeloni-usds, can you clarify?

projectcharter.md
ju-liem: I included some additional explanation in the Exit Plan section. 

I've been thinking of this particular charter as being tightly constrained to the initial scope of work through the end of September, and any additional on-going/follow-on work would be picked up in a separate document. This could very well not be the best way to approach this. I'm definitely open to shifting to another framing.

jcmeloni-usds:
I've made a bunch of edits and am going to merge as-is -- we can use the new version as a base to edit from, should burning additions or changes be desired...such as when some of the "to-dos" listed become isssues that we then complete.  I'll make issues out of the to-dos next.

This is still a work in progress, not a final document.  It also doesn't have to be perfect even when it's final, just something we can all agree to.

PR #23, opened 2018-04-26
Update README, charter

jcmeloni-usds:


jcmeloni-usds:
Thanks for fixing my typos! :)

PR #56, opened 2018-05-24
Set up test infrastructure

annekainicUSDS:
## Description
Install packages and set up configuration necessary for unit tests and linting to run properly. Include one sample test to ensure things are working, but will port the entire test suite in a separate PR.

## Additional information
I did not not include functionality for e2e tests yet because I'm not sure what kinds of e2e tests we'll want to create for the library. I think we should address that as a separate ticket.


package.json
```diff
@@ -48,15 +63,22 @@
...
+    "sinon": "^3.2.1",
+    "skin-deep": "^1.0.0"
   },
   "dependencies": {
     "babel-plugin-dynamic-import-node": "^1.2.0",
+    "blob-polyfill": "^2.0.20171115",
+    "choma": "^1.1.0",
     "downshift": "^1.22.5",
     "@department-of-veterans-affairs/react-jsonschema-form": "^1.0.0",
+    "eslint-plugin-import": "^2.7.0",
```
jordangov: Wondering why these two eslint deps are not in devDeps?

package.json
```diff
@@ -48,15 +63,22 @@
...
     "redux": "^3.5.2",
-    "redux-thunk": "^2.1.0"
+    "redux-thunk": "^2.1.0",
+    "sinon": "^3.2.1",
+    "skin-deep": "^1.0.0"
   },
   "dependencies": {
     "babel-plugin-dynamic-import-node": "^1.2.0",
+    "blob-polyfill": "^2.0.20171115",
+    "choma": "^1.1.0",
```
jordangov: Should this be in devDeps?

package.json
annekainicUSDS: Good question, short answer is I've been transferring a lot of the set up from the Vets.gov codebase and that's how these packages were installed, as regular dependencies. Looking into them, I think it's possible to just have them installed as devDependencies, so I'll change that for now and we can always switch back if for some reason an issue arises with that in the future.

jordangov: 👍

annekainicUSDS:
Can confirm the linter catches errors because there are currently many errors. However, I'm going to be doing a new port of the files from vets.gov, so soon those will all be fixed. Test also runs and passes (when I change the environment file, which will also be changed in the new port). Going to merge!

jcmeloni-usds:
thank you @jordangov :)

PR #63, opened 2018-05-30
New port from vets website

annekainicUSDS:
## Description

Redoing port from vets-website with latest changes. Also testing the integration of this port with this branch on vets-website: https://github.com/department-of-veterans-affairs/vets-website/tree/test-us-forms-system-integration.


annekainicUSDS:
Unit tests currently are not passing for 2 reasons:
1. There's scroll functionality that checks for `window.VetsGov.scroll`. This is set within the Vets.gov codebase. Obviously we don't want this library to be dependent on that. Replace the `window.VetsGov.scroll` with something specific to forms instead.
2. There's one reference to `__BUILDTYPE__`, which is a global variable set in Vets.gov. This reference comes from the component `SubmitButtons` and is used only to enable functionality that we wouldn't want to allow in production, but is useful in testing. In looking into this, I noticed that there's a lot of Vets.gov-specific content in `SubmitButtons`. Figure out a way to make `SubmitButtons` more generic, or replace it entirely with a generic component for the forms library and allow Vets.gov to continue to use their unique version.

annekainicUSDS:
Fixed the above 2 issues, unit tests are now passing!

PR #74, opened 2018-06-08
Move packages and update instructions

annekainicUSDS:
## Description

Some packages needed to be moved around, some to peerDependencies and some to devDependencies. Update getting started instructions to reflect these changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
   },
   "peerDependencies": {
     "@department-of-veterans-affairs/formation": "^0.8.0",
-    "vets-json-schema": "https://github.com/department-of-veterans-affairs/vets-json-schema.*"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -87,7 +90,6 @@
     "recompose": "^0.21.2",
     "redux": "^3.5.2",
     "reselect": "^2.5.4",
-    "uswds": "^1.6.3",
-    "vets-json-schema": "https://github.com/department-of-veterans-affairs/vets-json-schema.git#86263a396bfe2b13153797bac78c424a913b450d"
+    "uswds": "^1.6.3"
```
jbalboni: Probably could move this to peer dependencies as well.

annekainicUSDS: I can just make that change quickly now!

PR #83, opened 2018-06-18
Remove vets-json-schema as a peer dependency

dmethvin-gov:
## Description

vets-json-schema is used in the unit tests and already referenced as a devDependency. However, it as also listed as a peerDependency which is not necessary since consumers of this package aren't required to install it.

I get one "Pending" unit test,

>     - should fill in other formData when a suggestion is selected

Also this message:

> Warning: Failed prop type: The prop `onValueChange` is marked as required in `ErrorableCheckbox`, but its value is `undefined`.

These occur in master already though so I don't think they're related.

Before you hit Submit, make sure you’ve done whichever of these applies to you:

- [x] Run `npm test` and make sure the tests for the files you have changed have passed.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -10,12 +10,11 @@ If you are starting a completely new application from scratch, the easiest way t
 
 If you are installing the library into an existing application, you can follow these steps:
 
-1. Install the library: 
+1. Install the library:
 `npm install --save https://github.com/usds/us-forms-system.git`
 
 2. Install peer dependencies:
 - `npm install --save-dev @department-of-veterans-affairs/formation`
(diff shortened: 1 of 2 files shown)


annekainicUSDS:
LGTM! One other requested change: can you please update the [getting started docs](https://github.com/usds/us-forms-system/blob/master/docs/getting-started.md#2-install-in-an-existing-application) to not require installing vets-json-schema (step 2)?

And I can make a ticket for those messages in the unit tests.

dmethvin-gov:
Landed in 108614f2d61ea1a44c07d9ce791bd260f2d34d8b

PR #91, opened 2018-06-20
Move forms-related styles from VA design system

annekainicUSDS:
## Description

I did a high-level audit today of the CSS that is imported from the VA design system. I moved styles that were only used by the forms code into this repo, and am removing those styles from the VA design system entirely in [this PR](https://github.com/department-of-veterans-affairs/design-system/pull/125).

However, there are some styles that are not so easily moved over. The VA design system contains several components that are used by forms, but also used elsewhere (e.g., `ErrorableCheckbox`), and the styles for those components are needed both by us-forms-system and Vets.gov in general. I'm not sure what the best approach is yet for those styles, but figured I would tackle that in a separate PR anyway.


src/scss/styles.scss
```diff
@@ -37,11 +37,12 @@
 @import "node_modules/@department-of-veterans-affairs/formation/sass/base/b-mixins";
 @import "node_modules/@department-of-veterans-affairs/formation/sass/modules/m-dropdown";
 @import "node_modules/@department-of-veterans-affairs/formation/sass/modules/m-form-elements";
```
dmethvin-gov: Did you mean to leave this in? There's a local m-form-elements now.

annekainicUSDS: Yes, I meant to leave it in. Unfortunately, not all of the styles in the m-form-elements file from the design-system could be moved here, so I pulled out only the ones that can be for now.

dmethvin-gov: Ah, ok. It may turn out we want to refactor some of that into its own components if they're usable outside this repo by themselves.

annekainicUSDS: Yes, absolutely!

PR #89, opened 2018-06-19
Fix unit test messages

dmethvin-gov:
Eliminates warning message in unit tests

Partial fix for #84

- [x] Run `npm test` and make sure the tests for the files you have changed have passed.

PR #95, opened 2018-06-20
Create scss variables file

annekainicUSDS:
## Description

Create scss variables file to make it easier for users to override base styles.


lib/js/review/SubmitController.js
```diff
@@ -138,7 +138,7 @@ var SubmitController = function (_React$Component) {
         ),
         _react2.default.createElement(_PrivacyAgreement2.default, {
           required: true,
-          onChange: this.props.setPrivacyAgreement,
+          onChange: _actions.setPrivacyAgreement,
           checked: privacyAgreementAccepted,
```
dmethvin-gov: This looks like a result of me not doing a `npm run build` on PR #89 right? The CI will be doing that automatically but we may also need a commit hook so that the built files don't get out of sync with the original source.

annekainicUSDS: Yes, I think so. A commit hook is a good idea. I'll create a ticket for that.

annekainicUSDS:
So far things are working without importing any stylesheets from VA's design system! A few things left to do:
- Test on vets-website to make sure it doesn't break forms styles there
- Figure out why checkbox at the end of the form isn't being filled when clicked
- Remove commented out imports in styles.scss once I've confirmed everything is working as it should

annekainicUSDS:
Tested integration with vets-website. Things work well without importing these styles. However, importing these new styles is not working; however it's also not working on master right now. More changes will be needed to fix that, but trying to push that into this PR is too much. I'd like to reserve that for a separate PR. Vets.gov can still use the most recent version of us-forms-system without a problem. @dmethvin-gov do you want to take another look before it gets merged?

src/scss/styles.scss
```diff
@@ -34,14 +34,9 @@
 @import "node_modules/uswds/src/stylesheets/components/navigation";
 @import "node_modules/uswds/src/stylesheets/components/sidenav";
 
-@import "node_modules/@department-of-veterans-affairs/formation/sass/base/b-mixins";
-@import "node_modules/@department-of-veterans-affairs/formation/sass/modules/m-dropdown";
-@import "node_modules/@department-of-veterans-affairs/formation/sass/modules/m-form-elements";
 @import "node_modules/@department-of-veterans-affairs/formation/sass/modules/m-progress-bar";
```
annekainicUSDS: This is going to be removed in a separate PR.

Issue #68, opened 2018-06-06
Publish as npm package


annekainicUSDS:

jcmeloni-usds:
Exciting!

annekainicUSDS:
Does USDS have an npm organization? We probably want to publish it under that as opposed to a personal name.

jcmeloni-usds:
@nacingov thoughts?

dmethvin-gov:
It looks like the usds organization name is available, I can create it. They have the option of a free account with unlimited public repos or a $7/month account that allows private repos. I don't think we need private repos do we? We'll only use npm for our public open source projects?

jcmeloni-usds:
@dmethvin-gov We don't need a private repo -- anything we do organizationally will/should be open.  Go for it, unless @nacingov or @dnesting-usa call a halt to doing so (if they don't respond by Friday, just do it and I'll take any heat).

dnesting-usa:
This seems fine to me but @nacingov might have thoughts.

dmethvin-gov:
I'm a person of action so I created https://npmjs.com/org/usds with just free accounts for now. I suspect we could change it if we really want a paid account. A couple of questions:

**Management of the org:** At the moment there are no other npm accounts associated, I need to add some so it's not just me who can manage it. Publish rights for a specific project can be managed separately from the USDS org so perhaps @dnesting-usa and @nacingov for starters? There is [support for two-factor auth on npm](https://docs.npmjs.com/getting-started/using-two-factor-authentication) so I recommend that we all use it.

**Naming of projects:** We can either publish `us-forms-system` at the top public level or inside our org namspace as `@usds/us-forms-system`. If we're intending this to be for public consumption I'd recommend that we put at the top public level. Projects like `us-forms-system-starter-app` don't need to be published on npm since they're not meant to be put in another project's `package.json`. We *could* put them there but I'm not sure it facilitates anything.

dnesting-usa:
> I'm a person of action so I created

I don't care what Nacin says.  You are of our tribe.

jcmeloni-usds:
same

dmethvin-gov:
See #101

nacingov:
For the record, I agree with everything here.

dmethvin-gov:
Here's the way I see it working, ideally:

* Team agrees that we're ready to cut a release
* Someone submits a pull request that includes a new tag with the version number, e.g. `v1.2.3`, pointed to the HEAD of master
[...]

PR #101, opened 2018-06-22
Add npm publishing

dmethvin-gov:
Refactoring and adding support for publishing to npm.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,9 +1,13 @@
 version: 2
-jobs:
-  build:
+
+common: &common
     docker:
       - image: circleci/node:8.11.3-browsers
-    working_directory: ~/us-forms-system
+    working_directory: ~/repo
+
+jobs:
+  build:
+    <<: *common
     steps:
       - checkout # to working_directory
       - run: 'sudo npm install -g npm@6'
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -16,4 +20,31 @@ jobs:
...
+      - build:   # run for all branches and PRs
+          filters:
+            tags:
+              only: /.*/
+      - deploy: # run for only tagged (master)
+          requires:
+            - build
+          filters:
+            tags:
+              only: /^v.*/
```
annekainicUSDS: Will this cause a problem if for some reason we have a tag name that starts with a `v` but isn't a version number? should we make this regex more specific?

.circleci/config.yml
```diff
@@ -1,9 +1,13 @@
 version: 2
-jobs:
-  build:
+
+common: &common
     docker:
       - image: circleci/node:8.11.3-browsers
-    working_directory: ~/us-forms-system
+    working_directory: ~/repo
```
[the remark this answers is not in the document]
dmethvin-gov: This is the local directory in the docker image that CircleCI is using to run the tests. Originally I was thinking it was a good idea to use the project name, but inevitably I'll be copying this to other projects and it just creates another thing that needs to be edited before the file can be used elsewhere.

.circleci/config.yml
dmethvin-gov: In general I've only used git tags for versions in the master branch of repos, and that must be pretty common because all the CircleCI examples are assuming it too. You may have noticed the seemingly useless `/.*/` match everything in the `build filters tags` step above, it's needed to make this work right. Maybe you're getting the impression that I'm not a CircleCI fan, and you'd be right 😼, but all these CI systems seem to have crazy quirks.

dmethvin-gov:
I'm going to close this, as I think there may be a different way to do it that I'll discuss in #68.

PR #117, opened 2018-06-26
Use babel-cli to build; remove build artifacts before rebuild

dmethvin-gov:
I removed the script dir since we no longer need it. We can always add it back if there's something to put there.


package.json
```diff
@@ -8,7 +8,7 @@
     "test": "test"
   },
   "scripts": {
-    "build": "node script/build.js && npm run compile-sass && npm run autoprefix",
+    "build": "rm -rf lib/{js,css} && npx babel src/js --out-dir lib/js && npm run compile-sass && npm run autoprefix",
```
dmethvin-gov: I'm assuming that all the js and css files are expendable in `lib/`. If we don't remove them forcibly then old copies of removed files would remain around unless we removed them explicitly.

script/build.js
```diff
@@ -1,44 +0,0 @@
...
-      if (file.indexOf('.') === -1) {
-        if (!fs.existsSync(destinationPath)) {
-          fs.mkdir(destinationPath);
-        }
-
-        // Reset path to be the contents of the folder
-        const newPath = originalPath + '/';
-
-        // Recursively call function to go through files within subdirectory
-        createCompiledFiles(newPath, entryDestinationPath, startIndexOfString, compilationFunction);
```
annekainicUSDS: Bye, bye, recursive function 👋

package.json
annekainicUSDS: Looks like this was a feature request at some point, but seems like it didn't go anywhere... https://github.com/babel/babel/issues/5636

annekainicUSDS: I'm not really sure how this would work (if this belongs in the realm of this project or in a downstream project installing us-forms-system), but sourcemaps are really useful, and seems like there's a way to include them with babel-cli: https://babeljs.io/docs/en/babel-cli.html#compile-with-source-maps.

dmethvin-gov: I'll open a new ticket for that, it's a good idea. At minimum we'll needt to add the `--source-maps` flag and commit the resulting `.map` files.

PR #122, opened 2018-06-28
Add js sourcemaps

### LINKED ISSUE #119: Add source maps for JS files
As suggested in https://github.com/usds/us-forms-system/pull/117#discussion_r198317767 .

dmethvin-gov:
Fixes #119

Additional config changes may be required in the sample app as well, so that the maps survive the long and treacherous journey through the final compressed JS files there. I'll check that out later.


dmethvin-gov:
The browser is running a compiled version of the files that don't look like the source we edit. With a sourcemap, the browser can translate the location it's at in the JS or CSS file into the actual location in the original JSX or SCSS file that we edit, and show that instead.

There need to be separate map files because many of the the individual JS files are only included by an app if they need that functionality. So for example if an app doesn't use file fields on its form, it won't include [fileField.js](https://github.com/usds/us-forms-system/pull/122/files#diff-1a512c12c24bc24dddd9b5ba69ee012c) which is the compiled version of [fileField.jsx](https://github.com/usds/us-forms-system/blob/7869315d357a5cdc75a257abe519e07840991582/src/js/fields/FileField.jsx). But if it does need file fields, it will include the JS file and dev tools will use [fileField.js.map](https://github.com/dmethvin-gov/us-forms-system/blob/4639d1628108c54f1f2cea996c3c2924a0035a40/lib/js/fields/FileField.js.map). 

Inline maps can work okay for an app (vs a library) because the build step in the app decides whether the maps are needed or not (usually, development vs production). It's still very handy to have a map when debugging a production app so that you can tell what the original source looks like, so it's not necessarily true that you don't want a map for the production build. Sometimes that's when a map really saves your bacon, when you need to debug the live app.

However, it's even worse with a library. That's what #121 was about. The styles.css file went from about 150KB to 500KB when it used an inline map. Since you only need the map when you're actually inside dev tools and debugging, it's a big penalty to pay for the majority of the times when you have a debug build but aren't actually looking at the map. It also makes the library more error-prone to consume because the app that includes us-forms-system would need to strip out the inline maps somehow, or we'd have to publish two versions (production and debug), or we decide not to have maps at all. We publish both the JS and the JSX files so if we publish the map for each file as well it should be possible for an app writer to see the us-forms-system JSX which would make debugging easier.

PR #124, opened 2018-06-29
Add npm publishing

dmethvin-gov:
According to [this blog post](https://circleci.com/blog/publishing-npm-packages-using-circleci-2-0/), this will do what we want. I've tried testing on my local repo but it's super difficult and error-prone to do that so I'm going to put this up on the real repo. 

Note that the repo is currently on 1.0.0. Once we start publishing to npm we should follow semver which means we are likely to be increasing the major version relatively frequently if we make breaking changes, which seems inevitable in the near term. There is no problem with that but sometimes people see version numbers as a sign of library maturity (e.g., *"It's at version 5.1.3 so it must have been around for a while."*)

Once I've ensured this is all working properly I'll write up instructions for publishing. It's a bit tricky since we can't push directly to master because the tagged commit is in the PR and **must** land without being rebased/squashed and it should be at the head of master at the point where it's landed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,9 +1,13 @@
 version: 2
-jobs:
-  build:
+
+common: &common
     docker:
       - image: circleci/node:8.11.3-browsers
-    working_directory: ~/us-forms-system
+    working_directory: ~/repo
+
+jobs:
+  build:
+    <<: *common
     steps:
       - checkout # to working_directory
       - run: 'sudo npm install -g npm@6'
@@ -16,4 +20,31 @@ jobs:
             - ./node_modules
       - run: npm run lint
       - run: npm test
-      # See https://circleci.com/docs/2.0/deployment-integrations/ for deploy examples
+      - persist_to_workspace:
+          root: ~/repo
+          paths: .
+
+  deploy:
+    <<: *common
+    steps:
+      - attach_workspace:
+          at: ~/repo
+      - run: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > ~/repo/.npmrc
+      - run: npm publish
+
+workflows:
+  version: 2
+  build-deploy:
+    jobs:
+      - build:   # run for all branches and PRs
+          filters:
+            tags:
+              only: /v.*/
+      - deploy: # run for only tagged (master)
+          requires:
+            - build
+          filters:
+            tags:
+              only: /^v.*/
+            branches:
+              ignore: /.*/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,4 +42,4 @@ workflows:
             - build
           filters:
             branches:
-              only: master
+              only: production
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,17 +34,12 @@ jobs:
 
 workflows:
   version: 2
-  build-deploy:
+  build_and_deploy:
     jobs:
-      - build:   # run for all branches and PRs
-          filters:
-            tags:
-              only: /v.*/
-      - deploy: # run for only tagged (master)
+      - build
+      - deploy:
           requires:
             - build
           filters:
-            tags:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -16,4 +20,31 @@ jobs:
...
+      - attach_workspace:
+          at: ~/repo
+      - run: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > ~/repo/.npmrc
+      - run: npm publish
+
+workflows:
+  version: 2
+  build-deploy:
+    jobs:
+      - build:   # run for all branches and PRs
```
annekainicUSDS: Do we need to include steps like running our test suite and linter here? I see that's what USWDS is doing. Even though we run these steps before a PR can be merged, it seems like there could potentially be a risk of 2 PRs being merged in such a way that PR#2 could break a test, but CI wouldn't catch it in the PR review stage because it's being reviewed against master, and the test is something introduced by PR#1. Because we don't require rebasing on master before master, this kind of situation, while unlikely, is possible.

.circleci/config.yml
```diff
@@ -16,4 +20,31 @@ jobs:
...
+      - persist_to_workspace:
+          root: ~/repo
+          paths: .
+
+  deploy:
+    <<: *common
+    steps:
+      - attach_workspace:
+          at: ~/repo
+      - run: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > ~/repo/.npmrc
```
annekainicUSDS: Do we also need to add the `.npmrc` file?

dmethvin-gov: The `.npmrc` file has the token so we don't want to publish that to the repo. However, that does bring up another thing; npm will ignore everything in the `.gitignore` file but I think there may be a couple of other things we want to ignore like `.circleci` so I should add that. Thanks for asking!

annekainicUSDS: Ah gotcha. So where does the `.npmrc` file reside in order for this work? Sorry if these are basic questions!

.circleci/config.yml
dmethvin-gov: I have gotten dizzy trying to figure these scenarios out over the past week and make them less error prone. The way I've done it on other repos is to push a tagged commit directly to master. When you do that there's no risk of missing a commit to master because when you try to push, git will stop you since your local doesn't contain the newer commit.

Do you know the folks who are working on USWDS? I'd be interested in knowing a step-by-step on how they do releases because their CircleCI file is simpler.

annekainicUSDS: Yes! I can introduce you to some folks at USWDS. I'll send an email introduction.

.circleci/config.yml
dmethvin-gov: My specialty is just-in-time knowledge as well, didn't know CircleCI until a week ago and still don't feel like I know it. What's happening here is that we're checking out a "checkpointed" system image, pulling down the GitHub repo at master, and doing a clean install. If the tests pass we can do an `npm publish` but have to have credentials. Since that's secret and we don't want to let it be known, CircleCI stores the key securely and only makes it available in the environment. We have to echo that out to the `.npmrc` file before trying the publish.

Which brings up another point. At the moment the token configured in CircleCI is my account, one of the todos is to figure out how to get a token that isn't associated with a particular person, we probably should have an account something like "USDS-Publisher" on npm and use a token from that. But then we also need to store its password somewhere that appropriate people can log in if needed. I'll ask Maya about that too and see whtat USWDS is doing.

annekainicUSDS: great point!

dmethvin-gov:
I've updated the CircleCI setup to use `production` for npm releases and changed the [Release process wiki page](https://github.com/usds/us-forms-system/wiki/Release-Process) to reflect this. I still have some skeletal info in that page that I will fill in once I've been through the process once to know whether it tells the truth or not.

PR #130, opened 2018-07-02
Add missing date css

annekainicUSDS:
## Description

In building a sample form using the starter app, I noticed some css was missing to style date inputs appropriately. I've copied those styles in from the VA design system.

A larger question for us to answer at some later date is if we want to go back to using the USWDS default date input. On Vets.gov, we decided to deliberately diverge from how USWDS handles dates and build our own custom component (dropdowns for month and day instead of text fields), with custom styling. Do we want to go back to using the USWDS version of a date field?


dmethvin-gov:
Is the USWDS approach less useful than the vets.gov one? If so, I wonder if it should be suggested back to USWDS as a change. I'd be inclined to go back to the USWDS setup just to keep things simpler, but I suppose that has implications on whether we break vets.gov?

annekainicUSDS:
To be honest, the reason why Vets.gov diverged from the USWDS approach was not for a great reason. For the first ever form we built on Vets.gov, our stakeholder requested that we use dropdowns for month and day as opposed to text fields, which is what USWDS uses for those 2 fields, so that's what we did.

If we go back to the USWDS one, it will involve more work because the basic `DateWidget` will need to be changed, which will change it on Vets.gov automatically. I think it's worth having a more intentional discussion around this as opposed to just making this change outright, and perhaps chat quickly with the Vets.gov design team to see if they have discovered anything in their user research that would impact that decision. So I thought the quickest thing to do for now was to make this CSS addition and the larger decision around which version of the date field to use can be made separately.

annekainicUSDS:
Great idea, that would be a good use case!

PR #133, opened 2018-07-03
Don't assume Google Analytics is present

dmethvin-gov:
The forms system attempts to report form stats (errors or successful submits) via Google Analytics. The current code throws an error if `window.dataLayer` is missing. This change skips GA reporting if GA hasn't been loaded. That case can occur either because the app author elected to not use GA in their project or the user's browser has refused to load GA, for example it is using an ad blocker.


dmethvin-gov:
Sorry, I'm not used to having the built files committed! I guess I need to work on that commit hook. 😺

PR #136, opened 2018-07-03
Fix error message styling

annekainicUSDS:
## Description

Connected to https://github.com/usds/us-forms-system-starter-app/issues/9

The problem with error messages being displayed improperly came down to CSS. The reason why "Error" was being displayed before each message was because it was meant to only be read by screenreaders. I needed to add styling (copied from VA's design system) to hide these classes.

In the process of testing things I found another styling issue related to error messages. The date inputs included extra space in an error state, so copied those styles from VA's design system too to collapse that space.

Custom error messages are being displayed properly, so that aspect still works.


annekainicUSDS:
Yes, there's a chance there are other missing styles. When I removed references to VA's design system, I didn't want to copy over every stylesheet we were importing before, I tried to only pull in things actually needed. There have been some styles I missed, but I'd rather add them as we find we need them than import more than we need.

PR #148, opened 2018-07-10
Add IDEs and package managers to tools list

bernars-usa:
These were issues I ran into with my new machine, but @ju-liem confirmed she was also running into trouble with them, so it doesn't hurt to make the list really really complete.

@usds/forms-system for 👀

### DIFF
--- a/docs/getting-started/tools-for-getting-started-with-the-us-forms-system.md
+++ b/docs/getting-started/tools-for-getting-started-with-the-us-forms-system.md
@@ -1,47 +1,46 @@
 # Tools for getting started with the US Forms System
 
-Building a form using the US Forms System requires some minimum front-end development tools, as well as general knowledge of those tools. Where possible, we'll provide links to information about getting started with these tools.
+Building a form using the US Forms System requires some minimum front-end development tools, as well as general knowledge of those tools. Where possible, we've provided links to information about getting started with these tools.
 
 ### In this guide
 
-- [Command line interface (CLI)](#command-line-interface-cli)
-- [Integrated development environment (IDE)](#integrated-development-environment-ide)
-- [Package manager](#package-manager)
-- [JavaScript](#javascript)
-- [Node.js and npm](#nodejs-and-npm)
-- [React](#react)
-- [JSON Schema](#json-schema)
-- [react-jsonschema-form (rjsf)](#react-jsonschema-form-rjsf)
+- [General development tools and knowledge](#general-development-tools-and-knowledge)
+- [Laungauges and technologies](#languages-and-technologies)
+  - [JavaScript](#javascript)
+  - [Node.js and npm](#nodejs-and-npm)
+  - [React](#react)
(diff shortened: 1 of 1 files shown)


bernars-usa:
> I, for example, don't use an IDE, just a terminal and a text editor. I think some of these suggestions could be misleading.

This is definitely a fair point. What @ju-liem and I ran into was something more akin to "new machine syndrome" where in order to get the tools in the list installed, we had to get some more basic things like Xcode and Homebrew set up, especially because of the restrictions with the computers we're using. I wanted to find a way to provide the *most basic* information about how to get around those issues without actually prescribing a dev environment.

I reorganized what I added here based on a brief conversation with @jcmeloni-usds (https://github.com/usds/us-forms-system/pull/148/commits/c360ecef73053c414637f56246c4d2cc74d552bb)... what do you think? Still too much?

annekainicUSDS:
@bernars-usa thanks for taking another look at this, but I still believe this content doesn't belong in our repo. The issues you described running into are not related to our library, as you said, but just general issues in getting your machine set up for local development. I've never seen a framework or library include guidance on these kinds of issues in their docs before. What we have as it is (e.g., in terms of command line tool and JavaScript skills) is beyond anything I've seen in a library or framework's documentation.

Even if we were to include guidance on general local development stuff, I wouldn't include anything about using an IDE. Again, that's a very specific choice that a developer makes on how they want to develop locally, and I think it's beyond the bounds of our project to recommend how they write their code.

I'm also confused by you needing Xcode, and Homebrew should only be used to install Node.js, nothing else. Everything else (e.g., React, rjsf, etc.) just gets installed using npm. The way it's written now it makes it sound like Homebrew is used to install everything, and that shouldn't be the case.

I think we should remove the content related to IDEs and Xcode. We could suggest using Homebrew as a way to install Node, but that's all I would add. Homebrew is not required to install Node. It's just one way to install it, and has the additional benefit of being able to manage different versions of Node. But if someone already has the version of Node installed that's needed for this library, there's no need for them to also be using Homebrew. Here's how vets-website writes their recommendation on using Homebrew: https://github.com/department-of-veterans-affairs/vets-website#requirements.

Let me know if this makes sense!

bernars-usa:
> Let me know if this makes sense!

This makes sense. Thanks for hearing me out, @annekainicUSDS. 👍

annekainicUSDS:
Of course! I definitely want to make sure it's clear how to use this library. There might be a way to include some instructions that say things like "we're assuming you have a way to edit the code locally. if you need information about text editors and IDEs, see here", and include some useful links.

bernars-usa:
Yes, I think there are more changes to be made to this article for sure. We're slowly getting more feedback about these Getting Started docs, so let's keep an eye on that feedback over the next couple of weeks.

I do want to say that often with open source projects, there are a lot of assumptions about the tools and abilities of the user. If we're not explicit about what's required, we run the risk of accidentally limiting the reach of the project.

Issue #134, opened 2018-07-03
Document the ability to track form stats and errors via Google Analytics


dmethvin-gov:
It's in there but there's no docs that I can see about what is reported.

dmethvin-gov:
Added Sentry (sentry.io via the include of the raven-js package), not sure how that's configured or whether it should stay.

bernars-usa:
There's definitely a need for some more content around search engine analytics in general. Are there best practices we want to note? It's probably not a good idea to release a web form without analytics hooks, even if they're very basic for tracking pageviews.

Issue #128, opened 2018-07-02
Technical docs for how to build a form


annekainicUSDS:
Separate from the docs we have already, which are more detailed descriptions of the specific fields/widgets, etc., is there enough information for someone to understand the steps to building a form? A tutorial or recipe for different patterns, etc.

dmethvin-gov:
For background, the reasoning behind this is driven by a scenario like the upcoming hackathon where people are unlikely to enter via the top of the docs tree. We'd like to get them to success quickly by having them focus on the form.js file and the essential knowledge they'll need in order to build that. A lot of it can be accomplished via copy/paste until they reach the more complex customization scenarios.

bernars-usa:
This is definitely a gap in the current docs. The idea of "[Creating a form config file](https://github.com/usds/us-forms-system/blob/creating-forms/docs/building-a-form/creating-a-form-config-file.md)" and "[Common patterns for building a form](https://github.com/usds/us-forms-system/blob/creating-forms/docs/building-a-form/common-patterns-for-building-forms.md)" was to have the basic "Setting up your config file" and then "Customizing your config file" docs. What should be expanded on in those as far as you all can tell?

bernars-usa:
Also linking up #137 here, which @dmethvin-gov opened this morning. ✨

bernars-usa:
> We'd like to get them to success quickly by having them focus on the form.js file and the essential knowledge they'll need in order to build that. A lot of it can be accomplished via copy/paste until they reach the more complex customization scenarios.

Does this feel adequately addressed with #149 for now?

annekainicUSDS:
I think this ticket is complete to improving the docs we currently have. I still think we should explore creating a tutorial, similar to what [the React tutorial](https://reactjs.org/tutorial/tutorial.html) does. I created a separate ticket to explore that specifically: https://github.com/usds/us-forms-system/issues/191. Going to close this one in the meantime.

Issue #79, opened 2018-06-14
Enhance Documentation for Starter App


ju-liem:
As @jcmeloni-usds used the starter app to create a first pass at the H2A superform, she noted that additional examples for the form config and information for getting started with the starter app would make it (even) more useful (than it already is).

@jcmeloni-usds, can you add some of your thoughts/ideas to this issue when you have time?

bernars-usa:
I'd be thrilled to incorporate thoughts and ideas for doc updates.

jcmeloni-usds:
Apologies in advance for this coming out generally as stream-of-consciousness but here we go:

* I think it's super important that we emphasize some familiarity ahead of time is needed with your system from a development standpoint.  By this I mean something like "Must be familiar with CLI, Node, npm, build systems".  (puts on eng manager hat) To me, this is a minimal requirement for all front-end developers using modern tools, and is a reasonable starting point otherwise we'll be dealing w/ help requests for all the different system nuances around getting npm etc set up on one's machine.  
** that being said, on my fresh ubuntu machine I had zero issues installing node and npm and getting the starter app started -- maybe 7 minutes total and that includes a detour around ubuntu's default node installation to get to the latest version (this is an ubuntu 16.04 issue not a USFS issue)
* Yes, that all means that the "you must be familiar with..." list is going to be longer than the "how to start this" list.  ha. 
* We should finalize if we're going to tell people to fork or clone.  I think we settled on clone because that makes sense with how people will actually use it (e.g. not committing back to usfs starter app), but we may need to wait a little bit until the usfs npm package is poublished and the usfs starter app is rejiggered to import that rather than how it's working now
* Beyond installation, I generally started working w/ form.js this way -- which is NOT how most people would do it, I don't think, because they wouldn't know where all the vets.gov stuff lives.  Also, this was a super stupid way for me to get started and I should have known better
** had the usfs documentation up
** had the mozilla rjsf documentation up
** had the 1010EZ form in front of me
** had the js for the 1010EZ in front of me
** tried to match what I wanted to use with what I was seeing in the 1010EZ with what I was looking for in the definitions files
** got frustrated and reminded myself just to start small and build up from there and not try to do it the "right" way by reusing components and creating definitions for commonly-used things
* When I got out of my own way and did it the longer way to at least get something on the screen, my questions were around the following...while I could figure them all out, they seemed to be of the low-hanging-fruit type that it might be nice just to tell people so they don't get bogged down in their own brains (or maybe that's just me!)
[...]

PR #149, opened 2018-07-10
Add docs for creating a simple form

dmethvin-gov:
I'm not totally comfortable with this because it seems pretty extensive for a "simple" form. In particular:

* If you leave the `introduction` page out the browser crashes with a "too much recursion" error so it's not really optional. It seems like there's some issue with it redirecting to `/introduction` which doesn't exist. I expected that if the intro was not specified it would go directly to the first page of the form. I wonder if `confirmation` is also crash-inducing since we haven't gotten to it in the starter app.

* Even for a simple one-page form you end up having to create wrapper chapters and pages. I'm also not sure if review should be required on a simple one-page form. It's super easy to make mistakes when creating a deeply nested `formConfig` from scratch and get nonsensical errors because it looks like `formConfig` isn't checked against a JSON Schema itself. If you miss a level of nesting on something it really confuses the code.

* There is a `schema` `title` and a `uiSchema` `ui:title`, it isn't clear to me when to use either although I used the `uiSchema` ones here so I'd have *something* in that object. However, it seems like using `schema` `title` would be a *lot* less typing and redundancy for this case. There's a [similarly confused person](https://github.com/mozilla-services/react-jsonschema-form/issues/957) who made a ticket in react-jsonschema-form. This particular case makes me think it would be super handy to have the code create a default title from the name of the field, e.g. `email` becomes `Email`.

* The `page` `title` seems like it could be optional but the React component says that it is not. The React component throws a warning in the console based on `propTypes` defined.

* The `chapter` `title` isn't required (no warnings) but has to be specified in practice. If it's not, the review page will show an empty untitled box in the accordion. Most likely we should require it, unless the review page can also be optional (if so, I don't know how yet). 

* When you get to the review page, the review component throws a `propTypes` warning because the `trackingPrefix` is not specified, this is related to the undocumented functionality mentioned in #134. I don't think we want to make Google Analytics a requirement.

I suspect this will need some changes, either to the text here in the PR, the code, or both. Fire away!

### DIFF
--- a/docs/building-a-form/creating-a-form-config-file.md
+++ b/docs/building-a-form/creating-a-form-config-file.md
@@ -4,10 +4,62 @@ Your form is generated from a JSON Schema configuration file called `form.js`, a
 
 ### In this guide
 
+- [Building a simple form](#building a simple form)
 - [Example `form.js` file](#example-form.js-file)
 - [About the `schema` and `uiSchema` objects](#about-the-schema-and-uischema-objects)
 - [Configuring `uiSchema` using rjsf options](#configuring-uischema-using-rjsf-options)
 - [Configuring `uiSchema` using US Form System options](#configuring-uischema-using-us-form-system-options)
+
+### Building a simple form
+
+The US Forms System contains a `FormApp` component that you configure and control via a `formConfig` object. In the starter app, the file that exports this object is located in `/src/js/config/form.js` and included in `/src/components/Form.jsx`. where `FormApp`is used. When building an app from scratch you can define and include these components where it's convenient.
+
+The `formConfig` has several sections, that include, for example:
+* A `title` of the form;
+* A set of `chapters` and `pages` that group form fields for display on the screen;
+* A `uiSchema` with instructions about the way form elements should be displayed and labelled;
+* A `schema` with information about the data resulting from the elements once they have been filled;
+* Optional `introduction` and `confirmation` pages that can appear before and after the form.
+
+The remainder of this page has much more detail about these and other options.
+
+Here is a simple form that requests a name, phone number, and email address:
+
+```js
+import Introduction from '../components/Introduction.jsx';
+
+const formConfig = {
+  title: "Simple form",
+  introduction: Introduction,
+  chapters: {
+    onlyChapter: {
+      title: "Contact Information",
+      pages: {
+        onlyPage: {
+          path: "contact-info",
+          title: "",
+          schema: {
+            type: "object",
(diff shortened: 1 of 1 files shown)


dmethvin-gov:
Addresses concerns in #79 and #128, maybe others.

docs/building-a-form/creating-a-form-config-file.md
```diff
@@ -4,11 +4,63 @@ Your form is generated from a JSON Schema configuration file called `form.js`, a
...
 
+- [Building a simple form](#building a simple form)
 - [Example `form.js` file](#example-form.js-file)
 - [About the `schema` and `uiSchema` objects](#about-the-schema-and-uischema-objects)
 - [Configuring `uiSchema` using rjsf options](#configuring-uischema-using-rjsf-options)
 - [Configuring `uiSchema` using US Form System options](#configuring-uischema-using-us-form-system-options)
 
+### Building a simple form
+
+The US Forms System contains a `FormApp` component that you configure and control via a `formConfig` object. In the starter app, the file that exports this object is located in `/src/js/config/form.js` and included in `/src/components/Form.jsx`. where `FormApp`is used. When building an app from scratch you can define and include these components where it's convenient.
```
annekainicUSDS: I might rephrase the opening paragraph here. I feel like the first concepts someone should understand before building a form is that they describe the form in `formConfig` and they need to render the top-level React component, `FormApp`, somewhere. I might write:

"In order to build a form, you'll use the configuration file to specify everything your form needs: questions, how questions are grouped together, content on the page, data validation, etc. In [the starter app](https://github.com/usds/us-forms-system-starter-app), the file that exports this configuration object is located at `/src/js/config/form.js`.

Most of the time you'll just be working within the config file to describe the form you need and you won't be writing React components. However, there are 2 cases where you'll need to work with React components directly:
1. **Rendering the top-level React component**: US Forms System uses a top-level React component, `FormApp`, to pass the `formConfig` object, exported from your config file, down to the rest of the components that are rendered. In order for the React component hierarchy to properly render, you'll need to include `FormApp` in _your_ app's top-level component, and pass it the `formConfig` object as a prop. See [how the starter app does this](https://github.com/usds/us-forms-system-starter-app/blob/master/js/components/Form.jsx) for an example.
2. **Creating custom components**: There may be times where you'll want to create your own custom React components to include in the form config. Examples may include creating a custom page you want to render (e.g., an introduction page), creating a component for content (e.g., a warning message) you want to render within a page, or creating a new component to override one of our default ones. In these cases you'll be building React components from scratch."

Something like that?

PR #151, opened 2018-07-11
Use pre-push hook to ensure build artifacts are up to date

### LINKED ISSUE #97: Add git commit hook to run build
And perhaps also linting?

dmethvin-gov:
Fixes #97

This uses Husky 1.0 which makes it easy to define any of the git hooks right in the package.json file.

I chose a pre-push hook which should prevent someone from accidentally creating a PR that has files out of sync. We can use a commit hook instead, but that could get annoying since it requires a complete rebuild before each commit. As we add more checks for things in the build process (e.g., link checking) it's likely to get even longer.

I think we may want to have a larger discussion about whether we really want to push build artifacts up to the repo for every commit, as opposed to just doing it on `production` releases. If we don't, the diffs become a lot more readable/relevant but it does prevent people from pulling a hash directly to use it (which can sometimes be a good thing).

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,11 @@
     "test": "BABEL_ENV=test mocha --require test/config/unit-test-setup.js --require test/config/enzyme-setup.jsx --compilers js:babel-core/register ./test/**/*.spec.jsx",
     "webpack": "webpack"
   },
+  "husky": {
+    "hooks": {
+      "pre-push": "npm run build && git diff --quiet || echo '** NOT PUSHED: Uncommitted changes! **'"
+    }
+  },
   "repository": {
     "type": "git",
     "url": "git+https://github.com/usds/us-forms-system.git"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -17,6 +17,11 @@
     "test": "BABEL_ENV=test mocha --require test/config/unit-test-setup.js --require test/config/enzyme-setup.jsx --compilers js:babel-core/register ./test/**/*.spec.jsx",
     "webpack": "webpack"
   },
+  "husky": {
+    "hooks": {
+      "pre-push": "npm run build && git diff --quiet || echo '** NOT PUSHED: Uncommitted changes! **'"
```
annekainicUSDS: Hm, can we not have the pre-push script just run the build itself as opposed to just warning someone that it hasn't yet been run? It would just save a step if it runs the build automatically. What do you think?

annekainicUSDS: Oops! Totally misread this. This looks good to me!

PR #167, opened 2018-07-17
[Ready for review] Reorganize content on building a form

bernars-usa:
Addresses #139. From that issue:

> This topic has thus far been pared down quite a bit from what existed originally, but it's by no means optimal. ... It's possible some of the content needs to be pulled out into separate topics altogether.

Right now, I've pulled the Quick Start content from @dmethvin-gov into a separate file, and renamed a few others. I'll leave some comments explaining what I moved where.

🖥 [Rendered](https://github.com/usds/us-forms-system/blob/schema-uischema/docs/building-a-form/README.md#building-a-form)

PR #175, opened 2018-07-19
[Ready for review] Update main Docs readme and update code comment

bernars-usa:
This PR fixes a couple of small things:

- It updates links in the main Docs readme according to the current reality, which is something I overlooked before merging #167. 😅 
- It takes a stab at a more helpful code comment about the `keepInPageOnReview` property (based on discussion in https://github.com/usds/us-forms-system/pull/167#discussion_r203408856).


docs/building-a-form/about-the-schema-and-uischema-objects.md
```diff
@@ -230,8 +230,10 @@ The us-forms-system code includes additional `uiSchema` functionality not found
       };
     },
 
-    // Use this when you have an array field that should not be pulled out of the page its
-    // in and shown separately on the review page.
+    // By default, array fields that are displayed on a single page in a form, such as
+    // information for multiple dependents, are displayed as separate pages or sections on
```
annekainicUSDS: Really nitpicky, but I might reword this to "are display in a separate section on the review page." Only because the way it works is all the info in the array field is displayed together, and the wording now makes it sound like each item within the array gets its own section, so just wanted to clarify that a bit!

bernars-usa: Does that mean this:

> To keep the information in a single section on a review page, set this property.

... should be something more like:

"To keep all information together on a review page, set this property."

I'm having a hard time deciphering the difference between when the property is set and not set according to your description.

annekainicUSDS: Sorry that wasn't clear. I would keep the wording "To keep information in a single section on a review page, set this property to `true`", just adding "to `true`" at the end. I think that language is the most specific about what's going on. The property only needs to be set to `true` if you want to keep the form fields together in one section; otherwise you don't need to set it all (setting it to `false` is the same as not setting it at all).

bernars-usa: Got it! Thanks for clarifying.

Issue #178, opened 2018-07-19
Cross-link between conceptual content and reference content about widgets and components


bernars-usa:
There are mentions of widget and component usage around the docs, and there are now reference docs about those widgets and components. It would be great to bring those closer together by cross-linking.

This came up in a discussion between @dmethvin-gov and @annekainicUSDS in https://github.com/usds/us-forms-system/pull/174#discussion_r203715156.

bernars-usa:
Along with this work, I'm also going to divide the widgets and definitions into two separate topics. We discussed this in https://github.com/usds/us-forms-system/pull/167#discussion_r203390197.

annekainicUSDS:
I think this has been complete with https://github.com/usds/us-forms-system/pull/182 and https://github.com/usds/us-forms-system/pull/184. @bernars-usa can you confirm that we can close this ticket?

bernars-usa:
Yes! This is complete.

PR #182, opened 2018-07-20
[Ready for review] Separate widgets and definitions into two separate reference topics

bernars-usa:
Addresses https://github.com/usds/us-forms-system/issues/178. The first part of this work is to:

- Separate the existing combined widgets and definitions topic into two separate topics
- Turn the table with rows of widgets into topic sections (so they can be linked to)

The definitions have brief descriptions, but the widgets don't. I'll need some help there.


docs/building-a-form/common-widgets.md
```diff
@@ -0,0 +1,78 @@
...
+- [`DateWidget`](#datewidget)
+- [`EmailWidget`](#emailwidget)
+- [`PhoneNumberWidget`](#phonenumberwidget)
+- [`RadioWidget`](#radiowidget)
+- [`SelectWidget`](#selectwidget)
+- [`SSNWidget`](#ssnwidget)
+- [`TextWidget`](#textwidget)
+- [`YesNoWidget`](#yesnowidget)
+
+### `ArrayCountWidget`
```
bernars-usa: @annekainicUSDS - For each of these, it would be great to have a short, useful description of the widget, similar to those provided for the definitions (see https://github.com/usds/us-forms-system/pull/182/files#diff-15372f86bdceb3c1bb7f70db64ca8471R32). I know some of these are described in "[Available form features...](https://github.com/usds/us-forms-system/blob/master/docs/building-a-form/available-form-features-and-usage-guidelines.md)," but the best wording is tripping me up a bit. Really appreciate your help!

docs/README.md
```diff
@@ -8,7 +8,8 @@
   - [About the `schema` and `uiSchema` objects](building-a-form/about-the-schema-and-uischema-objects.md)
   - [Quick Start: Example `form.js` file](building-a-form/quick-start-example-formjs-file.md)
   - [Creating a form config file](building-a-form/creating-a-form-config-file.md)
-  - [Adding widgets and definitions](building-a-form/adding-widgets-and-definitions.md)
   - [Available form features and usage guidelines](building-a-form/available-form-features-and-usage-guidelines.md)
+  - [Common widgets](building-a-form/common-widgets.md)
```
bernars-usa: Are these titles useful? We now have three topics called "Common..." and it's concerning me. 😕

docs/building-a-form/common-widgets.md
annekainicUSDS: Of course! Should I add this to the table or below as regular content?

bernars-usa: Below as regular content would be excellent, sort of in the fashion of a term/definition pair. The format in the definitions topic seems nice and succinct:

```markdown
### Current or past dates

The common date field with current or past validation set.
```

annekainicUSDS: Okay sounds good, I'll create a PR into this one

docs/README.md
dmethvin-gov: I suppose this particular one should be called Available Widgets but the others seem good as Common

docs/building-a-form/common-definitions.md
```diff
@@ -171,7 +142,7 @@ A function that takes a full name `schema` and sets its required field list to `
 - `schema`: Yes
 - Function or object: Function
 
-##### Person ID
+### Person ID
 
 A function that allows the user to input a form of personal identification. For example, the form can ask for a social security number and include a checkbox to specify that the user doesn't have one, allowing them to instead input a Veterans Affairs file number.
```
dmethvin-gov: Is *input* verbable nowadays?

docs/building-a-form/common-widgets.md
```diff
@@ -0,0 +1,78 @@
+# Common widgets
+
+These common widgets are included in the us-forms-system by default. Set these widgets in a config file while building your form.
+
+Some widgets are associated with particular schema types or formats. There are additional widgets available, but the us-forms-system uses definitions they're included in, rather than just the field. To use a widget, set `ui:widget` for a field to the name of the widget. Widgets are located in [/src/js/widgets](../../src/js/widgets).
+
+Some widgets are passed as a string (e.g., `'ui: widget': 'date'`), while others are passed as the component itself (e.g., `'ui:widget': CurrencyWidget`).
```
dmethvin-gov: ...passed *as a reference to* the component...

@annekainicUSDS I think in most cases these components are actually functions, right? So the examples below would need to look like: `'ui:widget': CurrencyWidgetUI('Total benefit')` and somewhere above there you would need `import CurrencyWidgetUI from 'us-forms-system/lib/js/definitions/currency'` to define the function.

docs/building-a-form/common-definitions.md
bernars-usa: As long as people understand what this means I'm fine with it. 🤷‍♀️

bernars-usa: Ugh, I'll change it to "type".

bernars-usa:
This is ready for a final review. Thanks so much for the feedback (and writing!) so far.

docs/building-a-form/available-widgets.md
```diff
@@ -0,0 +1,100 @@
+# Available widgets
+
+These widgets are included in the us-forms-system by default. Set these widgets in a config file while building your form.
+
+Some widgets are associated with particular schema types or formats. There are additional widgets available, but the us-forms-system uses definitions they're included in, rather than just the field. To use a widget, set `ui:widget` for a field to the name of the widget. Widgets are located in [/src/js/widgets](../../src/js/widgets).
+
+Some widgets are passed as a string (e.g., `'ui: widget': 'date'`), while others are passed as a reference to the component itself (e.g., `'ui:widget': CurrencyWidget`).
+- Widgets from the [react-jsonschema-form library](./about-the-us-forms-system-library#understanding-react-jsonschema-form-rjsf) include string mappings.
+- Widgets created specifically for US Forms System do not have these mappings, and therefore must be specified by passing the component for the widget directly to the config.
```
dmethvin-gov: Now I'm starting to wonder about splitting up the widgets and definitions. For most of the `uiSchema`-only widgets I think users are most likely to include them via the `js/definitions` that  exist rather than using `'ui:widget'`: USFormsWidgetName`?

If that's the way they plan to do it, we need to explain here that they need to import the definitions in JavaScript near the top of their file:
```js
import CurrencyWidgetUI from 'us-forms-system/lib/js/definitions/currency';
```
Then use the widget by calling it to add all the `uiSchema` definitions:
```js
uiSchema: {
  ...
  monthlyWages: CurrencyWidgetUI('Monthly wages')
  ...
}
```
People *do* have the option of including the widget directly as defined here, but even then they would need to `import` the widget [like the definitions do](https://github.com/usds/us-forms-system/blob/master/src/js/definitions/currency.js). Note that the definitions (like this one) may do quite a few things so trying to explain manual usage when there are definitions may not be productive. 

@annekainicUSDS can you sanity-check what I'm saying on this? I think the **Usage:** instructions need to be modified to make them clear if this is the case.

docs/building-a-form/common-widgets.md
dmethvin-gov: I guess what I'm seeing on forms like the [vets.gov 1990](https://github.com/department-of-veterans-affairs/vets-website/blob/master/src/applications/edu-benefits/1990/config/form.js) is that they aren't referencing any widgets at all directly, although they do mention `ui:widget` for the RJSF ones like `radio`. The thing about the non-builtins is that the definition files often have several lines and reference other things like a special review widget.

docs/building-a-form/available-widgets.md
bernars-usa: I've got some guidance here, could use some help with similar guidance for widgets: https://github.com/usds/us-forms-system/pull/182/commits/97e13c77ce19a99de5c9e60a86424331e18aa19e

annekainicUSDS: Added some comments on the above commit!

dmethvin-gov: For widgets, the similar text would be something like:
---
To include a widget, import the widget at the top of the file:
```js
import CurrencyWidget from 'us-forms-system/lib/js/widgets/CurrencyWidget';
```
Then and set the `ui:widget` field to the imported widget name:
```js
uiSchema: {
  ...
  'ui:widget': CurrencyWidget,
  ...
}
```

bernars-usa: Here's that information: https://github.com/usds/us-forms-system/pull/182/commits/bb892ab32d3c1e52941ecdf592543dd8665772bb

bernars-usa:
This is ready for another round of review! Since our last episode, we added guidance on using widgets and definitions. Take a look and let me know if this looks ready!

docs/building-a-form/available-widgets.md
```diff
@@ -0,0 +1,112 @@
+# Available widgets
+
+These widgets are included in the us-forms-system by default. Set these widgets in a config file while building your form.
+
+Some widgets are associated with particular schema types or formats. There are additional widgets available, but the us-forms-system uses definitions they're included in, rather than just the field.
+
+Widgets from the [react-jsonschema-form library](./about-the-us-forms-system-library#understanding-react-jsonschema-form-rjsf) include string mappings. Widgets created specifically for US Forms System do not have mappings, and therefore must be specified by passing the component for the widget directly to the config. To include such widgets, import the widget at the top of the file:
```
bernars-usa: @dmethvin-gov - Here's where I pared down information quite a bit, since it had gotten more specific. Let me know what you think.

docs/building-a-form/README.md
```diff
@@ -12,14 +12,18 @@ Use this example `form.js` file to build a basic form.
...
 
 These form features are available in the US Forms System library. We've provided information about how to implement them in your form.
 
+### [Available widgets](available-widgets.md)
+
+These common widgets are included in the us-forms-system by default. Set these widgets in a config file while building your form.
+
+### [Common definitions](common-definitions.md)
+
+These definitions address form field patterns that are more complex than widgets and include common label text, validation, or field components.
```
annekainicUSDS: I think this description is slightly inaccurate; it's not that definitions are more complex than widgets, but rather they accomplish something completely different. Widgets are simply used to render the HTML form element for a given field. Widgets don't know anything about the specific field they are rendering. They don't know anything about the text being rendered, the specific data that field is collecting, the context in which that question is being asked, etc. Their only job is to figure out what kind of HTML form element to render, sometimes with some additional functionality to make sure the data that gets entered into that field is handled properly.

Definitions are chunks of the config that have been extracted into a separate file because that chunk of config might be used in more than one place and it's easier to import a common piece of code from a separate file than to write the same piece of config multiple times. For example, you might have multiple full name fields throughout the form. By extracting that chunk of the form config into a separate file, you can more easily use it in multiple places. But if you don't want to use definitions, you can also just write everything to the form config directly. Definitions know about the specific way that question is going to asked: what is the label text? What is the error message? When is that question hidden? Are there additional CSS classes we want to pass to that field to format it differently?

Hopefully that explanation is helpful! Given all of that, I would rewrite this little blurb to be:

"Definitions are chunks of the form config for specific types of questions that can be dropped into the main form config where needed. They are usually used only for `uiSchema`, and include things like label text, validation functions, error messages, and which widget to render."

docs/building-a-form/available-widgets.md
annekainicUSDS: Do we want to include information from my comment above about the purpose of widgets, which is just to figure out which type of HTML form element to render?

docs/building-a-form/common-definitions.md
```diff
@@ -1,40 +1,29 @@
...
 
-These definitions address form field patterns that are more complex than widgets and include common label text, validation, or field components. Definitions are located in [/src/js/definitions](../src/js/definitions).
+Then, call it to add all the `uiSchema` definitions:
 
-- Simple definitions are provided as `schema` and `uiSchema` objects that you can import and overwrite to customize.
-- More complex definitions are functions that require certain parameters.
+```js
+uiSchema: {
+  ...
+  monthlyWages: currencyUI('Monthly wages')
```
annekainicUSDS: For this specific example, I might add text below it to clarify what this is:

"In this example, the definition is a function that takes the title for that field, which is used to populate the `'ui:title'` property in `uiSchema`."

docs/building-a-form/common-definitions.md
```diff
@@ -171,16 +160,16 @@ A function that takes a full name `schema` and sets its required field list to `
 - `schema`: Yes
 - Function or object: Function
 
-##### Person ID
+### Person ID
```
annekainicUSDS: This is pretty specific to Vets.gov, so perhaps we should consider removing it from our library?

docs/building-a-form/creating-custom-fields-and-widgets.md
```diff
@@ -17,20 +17,20 @@ rjsf passes all field and widget components to `SchemaField` (and most other com
...
+- [`EmailWidget`](available-widgets.md#emailwidget)
+- [`CheckboxWidget`](available-widgets.md#checkboxwidget)
+- [`RadioWidget`](available-widgets.md#radiowidget)
 - `TextareaWidget`
 
 The us-forms-system uses these custom fields and widgets:
 
-- `YesNoWidget`
+- [`YesNoWidget`](available-widgets.md#yesnowidget)
 - `AddressField`
```
annekainicUSDS: Do we need this? Not sure what it's referring to. There's a `definition` for address fields, but it doesn't seem to fit with this list, because these are all widgets.

docs/building-a-form/available-widgets.md
bernars-usa: Definitely. Will add. 👍

docs/building-a-form/README.md
bernars-usa: This reminds me that I also intended to incorporate the information from your comment in https://github.com/usds/us-forms-system/pull/182#discussion_r204173108:

> Widgets are React components that simply return specific HTML form elements. Definitions are chunks of schema (either regular schema or uiSchema) that can be plugged into the overall form config that describe aspects of the form.

I appreciate the careful thought about defining the difference between widgets and definitions. It's going to help a lot.

docs/building-a-form/common-definitions.md
bernars-usa: Do you mean update the description of Person ID to remove the VA file number reference, or remove the definition altogether? If the latter, I'll wait to remove the docs until the definition is removed.

docs/building-a-form/creating-custom-fields-and-widgets.md
bernars-usa: I can remove this if it's incorrect... it was in the docs already. Same goes for everything else here; it's all just taken from the legacy docs.

docs/building-a-form/common-definitions.md
annekainicUSDS: Ah I was thinking removing the definition altogether, which I agree we can do as a separate PR, so nothing to do here for now.

docs/building-a-form/creating-custom-fields-and-widgets.md
annekainicUSDS: Yep, I figured! All the work you're doing on reorganizing content is helping bring to light existing issues in the docs, so it's super helpful :)

docs/building-a-form/available-widgets.md
bernars-usa: @annekainicUSDS: The short description from the README is also included at the top of this page, so it's now there. Is that ok? Is there anything I should modify in this particular paragraph?

bernars-usa:
Ok! Ready for final final final review!

docs/building-a-form/common-definitions.md
dmethvin-gov: That makes sense. I was mentioning to @bernars-usa that this is just by convention and it would be possible for others to define these items with more than one arg or a multi-property object as an argument. Even the definitions we create take different numbers and orders of arguments so you have to look them up in the code at the moment.

docs/building-a-form/common-definitions.md
```diff
@@ -153,7 +142,7 @@ The common date field, excluding day field, with basic validation.
 - `schema`: No
 - Function or object: Function
```
dmethvin-gov: Probably best for a second PR, but each one that says it's a function should specify what arguments it takes.

bernars-usa: That seems important, yes. If you have time, could you open a PR to add those?

dmethvin-gov: yeah, I'll do that today so we have it before the hackfest!

PR #183, opened 2018-07-20
Fix the example for conditionally hiding fields

dmethvin-gov:
This is simplified from the big form I'm building.

### DIFF
--- a/docs/building-a-form/common-patterns-for-building-forms.md
+++ b/docs/building-a-form/common-patterns-for-building-forms.md
@@ -153,7 +153,7 @@ To create a block of text with no fields that follow, create an empty view objec
 
 Sometimes fields in a form are siblings to others, but should be hidden conditionally. For example, this schema defines a field named `employed` that is rendered as a checkbox the user can check. If this field is false (the checkbox is not checked) there is no need to display the `jobStartDate` or `monthlyWages` fields:
 
-```json
+```js
 schema: {
   type: 'object',
   properties: {
@@ -196,7 +196,7 @@ uiSchema: {
 }
 ```
 
-The hideIf function is passed a copy of the current formData in order to determine the condition upon which the fields are shown. In this example, it will hide the fields unless employed is true.
+The `hideIf` function is passed a copy of the current `formData` in order to determine the condition upon which the fields are shown. In this example, it will hide the fields unless `employed` is `true`.
 
 Objects from the form config with names that start with `view:` are not passed to the back-end, but fields within those `view:` objects are passed to the back-end by including them in the parent object. In this example, the fields `jobStartDate` and `monthlyWages` would be included in the same object as `employed` and `otherMonthlyIncome`, while the field for `view:JobInformation` would be filtered out.
 
--- a/docs/building-a-form/common-patterns-for-building-forms.md
+++ b/docs/building-a-form/common-patterns-for-building-forms.md
@@ -151,64 +151,52 @@ To create a block of text with no fields that follow, create an empty view objec
 
 ### Conditionally hiding a group of fields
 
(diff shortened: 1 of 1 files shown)


docs/building-a-form/common-patterns-for-building-forms.md
```diff
@@ -151,64 +151,52 @@ To create a block of text with no fields that follow, create an empty view objec
...
+      }
+    },
+    otherMonthlyIncome: { type: 'string' }
+  }
+},
+uiSchema: {
+  employed: { 'ui:title': 'I am employed' },
+  'view:jobInformation': {
+    'ui:options': {
+      hideIf: (formData) => !formData.employed
```
annekainicUSDS: Something I'm not totally clear on is the difference between `hideIf` and `expandUnder`. I think from reading [this function](https://github.com/usds/us-forms-system/blob/master/src/js/state/helpers.js#L98) that they are used to set 2 different `uiSchema` properties: `hideIf` sets `'ui:hidden'` and `expandUnder` sets `'ui:collapsed'`. Functionally, though, they seem to do the same thing. I think there are some benefits to using `expandUnder` if you want [to style](https://github.com/usds/us-forms-system/blob/master/docs/building-a-form/common-patterns-for-building-forms.md#styling-expanded-or-collapsed-fields) the expanded fields in a certain way, but otherwise I don't know if there's a big difference?

docs/building-a-form/common-patterns-for-building-forms.md
```diff
@@ -151,64 +151,52 @@ To create a block of text with no fields that follow, create an empty view objec
...
+    jobStartDate: { 'ui:title': 'Job start date' },
+    monthlyWages: { 'ui:title': 'Monthly wages' },
   },
-  other: { ... }
+  otherMonthlyIncome: { 'ui:title': 'Other monthly income' }
 }
 ```
 
-From this, the fields in the `view:sponsorServiceOptions` object are moved up one level and sent alongside `dic` and `chapter31`. The back end doesn't see objects with names that start with `view:`, but it gets all fields inside those objects.
+The back end doesn't see objects with names that start with `view:`, but it gets all fields inside those objects because the form processing includes them in the parent object (in this example, the one containing `employed` and `otherMonthlyIncome`). The `hideIf` function is passed a copy of the current `formData`, in this case it will hide the fields unless `employed` is `true`.
```
annekainicUSDS: The wording here was a bit confusing to me, perhaps this is clearer?

"The `hideIf` function is passed a copy of the current `formData` in order to determine the condition upon which the fields are shown. In this example, it will hide the fields unless `employed` is `true`.

Objects from the form config with names that start with `view:` are not passed to the back-end, but fields within those `view:` objects are passed to the back-end by including them in the parent object. In this example, the fields `jobStartDate` and `monthlyWages` would be included in the same object as `employed` and `otherMonthlyIncome`, while the field for `view:JobInformation` would be filtered out."

One thing we may want to answer is whether fields that are conditionally shown are always included in the final data object that is submitted to the back-end, or only included if they are shown. Looking at the code it seems like fields that are conditionally shown are included; I don't see any logic that filters those out. It's only complete pages that are filtered out. But it seems like it would make more sense if fields that are never shown to the user are also filtered out from the final submission? I need to do more digging into this.

docs/building-a-form/common-patterns-for-building-forms.md
dmethvin-gov: Honestly I don't know, the [logic looks nearly identical](https://github.com/usds/us-forms-system/blob/4d69ee604610de969e3d8d80f62fafabf87aa428/src/js/state/helpers.js#L113-L129) and [we consider them both hidden](https://github.com/usds/us-forms-system/blob/4d69ee604610de969e3d8d80f62fafabf87aa428/src/js/state/helpers.js#L9-L11). Other than that code I don't see anything that uses or examines it, even in RJSF?

Issue #171, opened 2018-07-18
Provide documentation around what is/what is not included in the library re: data persistence


afeld:
In the documentation and examples, I didn't see any recommendations around how to handle persistence, or even retrieval of the form data. Haven't actually poked around at the code to try and retrieve it, but any tips that could be added to the docs?

afeld:
Ah! Now seeing the mention of Redux in [the documentation](https://github.com/usds/us-forms-system/blob/master/docs/building-a-form/about-the-us-forms-system-library.md#how-us-forms-system-uses-rjsf), and the state structure in the [starter app](https://github.com/usds/us-forms-system-starter-app):

<img width="841" alt="screen shot 2018-07-17 at 10 48 36 pm" src="https://user-images.githubusercontent.com/86842/42856581-d6f3efa4-8a13-11e8-8671-ec05f4f1dbf3.png">

That might be a good thing to elaborate on in the docs, and then maybe just reference other documentation around how to persist data from Redux generally.

afeld:
Was also looking at [Redux Form](https://redux-form.com/) - curious how much overlap that has here.

annekainicUSDS:
Hi @afeld, thanks for this feedback! I agree, our docs could be more explicit about how we're using Redux. We're actively working on reorganizing and improving our docs right now, so this would be a good thing to include in that work. I will add this to our project board to keep track of.

jcmeloni-usds:
I failed to write anything about this.  It would be worth writing a brief entry in the library docs itself, maybe a paragraph or so, about current state and potential future state.

annekainicUSDS:
Add a link in our docs of the VA's implementation of retrieving form data with a brief description of how it works.

dmethvin-gov:
The VA pages wrap the form with [RoutedSavableApp](https://github.com/department-of-veterans-affairs/vets-website/blob/8042318adfb2130caebc9771373f4ab936e16784/src/platform/forms/save-in-progress/RoutedSavableApp.jsx) which provides data persistence and things like the "Are you sure you want to leave?" prompt if they try to navigate away from the page.

PR #187, opened 2018-07-24
[Ready for review] Edit conceptual info

bernars-usa:
This is a branch off of #186, which is where @annekainicUSDS took a huge step in improving the conceptual information about the library. This PR just fine-tunes that work a bit.


docs/README.md
```diff
@@ -12,6 +12,6 @@
   - [Available widgets](building-a-form/available-widgets.md)
   - [Common definitions](building-a-form/common-definitions.md)
   - [Common patterns for building forms](building-a-form/common-patterns-for-building-forms.md)
-  - [Creating custom fields and widgets](building-a-form/creating-custom-fields-and-widgets.md)
-- [How the library works](how-the-library-works/README.md)
```
annekainicUSDS: One thought about removing this topic on "How the library works": I was imagining other topics could go in here in the future. For example, in order to address [this ticket](https://github.com/usds/us-forms-system/issues/171), I was going to add a topic on how USFS uses Redux, and I assumed it would fit into a "How the library works" category, but curious on your thoughts here.

docs/customizing-the-library/creating-custom-fields-and-widgets.md
```diff
@@ -1,5 +1,7 @@
 # Creating custom fields and widgets
 
+You can customize the base library in order to build web-based forms using React and the JSON Schema standard.
```
annekainicUSDS: I think perhaps the more specific use case here is: "You can customize the base library in order to satisfy the unique requirements of your particular form." Or something like that. Not that it just lets you build web-based forms, because the base library lets you do that, but so that you can do specific things that aren't required in the base library. It's a little reductive, but at least it's more accurate. Let me know if that makes sense!

bernars-usa: Yeah, thank you, I tried to bulk this up but was going to ask you for more info. This is great!

docs/README.md
bernars-usa: I can help think through where to put that information. We really need these three different maps (sections) for these distinct audiences:

- People just introducing themselves to the library (Getting started)
- People who have everything set up and are building basic forms with no customizations (Building a form)
- People who are not only building forms, but building customizations on top of the library (Customizing the library)

These are what equate to "Beginner, Intermediate, and Advanced." Users do not self-identify into those groups accurately, so you have to be very explicit about the use cases.

bernars-usa:
Ok, this is ready for review! I moved some things around to get the new "Customizing the library" map topic, and did a copy edit of the new content @annekainicUSDS added (you can see that on it's own in https://github.com/usds/us-forms-system/pull/187/commits/39c9b7c490d82ca9dc6a279760a795cb06e9da66). Let me know if you have any questions!

docs/README.md
annekainicUSDS: I guess to me "Customizing the library" feels like a topic within a larger category of advanced things someone might need to know about potentially. For example, I can foresee future topics within an advanced category related to adding features for prefill and save in progress or how the state model works.

bernars-usa: > I can foresee future topics within an advanced category related to adding features for prefill and save in progress or how the state model works.

Definitely. We're eventually going to want to break up "Building a form" more, depending on future improvements to the library.

docs/customizing-the-library/creating-custom-fields-and-widgets.md
```diff
@@ -8,7 +10,7 @@
 
 ### How us-forms-system uses rjsf
 
-The us-forms-system code uses rjsf to render form fields, but it builds a scaffolding on top of it to support multi-page forms and our common form patterns. Additionally, us-forms-system uses rjsf to create a form configuration spec that allows devs to specify the structure of one of our multi-page forms.
+The us-forms-system code uses rjsf to render form fields, but it builds a scaffolding on top of it to support multi-page forms and our common form patterns. Additionally, us-forms-system uses rjsf to create a form configuration spec that allows developers to specify the structure of a multi-page form.
```
dmethvin-gov: The "our common form patterns" supposes a particular use case so I'd pull out "our".

docs/README.md
annekainicUSDS: Okay that makes sense! Sounds good :)

PR #188, opened 2018-07-24
[WIP] More detail on use of definitions

dmethvin-gov:
I skipped the `address.js` one for the moment because it's super crazy to explain, but is the format of these others okay? I think an example will be useful since it can be copy/pasted. I tried to de-emphasize the Source since that is not the actual code they include, although the link will be helpful since it contains the definitions and for ones like `address.js` it is super difficult to use without reading the source because of all the options it has.

### DIFF
--- a/docs/building-a-form/common-definitions.md
+++ b/docs/building-a-form/common-definitions.md
@@ -68,11 +68,10 @@ A common type-ahead widget that lets a user type in values and narrow down a lon
 The common electronic funds transfer (EFT) account information field that collects account type (checking or savings), bank account number, and bank routing number. Use these definitions as the uiSchema for the field that represents the account number.
 
 ```js
-import { uiSchema as bankAccountUI }
-  from 'us-forms-system/lib/js/definitions/bankAccount'
+import bankAccountUI from 'us-forms-system/lib/js/definitions/bankAccount';
 
 uiSchema: {
-  eftinfo: bankAccountUI;
+  eftinfo: bankAccountUI
 }
 ```
 Source: [/src/js/definitions/bankAccount.js](../../src/js/definitions/bankAccount.js)
@@ -82,11 +81,10 @@ Source: [/src/js/definitions/bankAccount.js](../../src/js/definitions/bankAccoun
 Formats and validates a US currency field. The display includes a leading `$` character. Call this exported function and pass it the label to be used on the field.
 
 ```js
-import { uiSchema as currencyUI }
-  from 'us-forms-system/lib/js/definitions/currency'
+import currencyUI from 'us-forms-system/lib/js/definitions/currency';
 
 uiSchema: {
-  payments: currencyUI('Total Payments');
+  payments: currencyUI('Total Payments')
 }
 ```
 Source: [/src/js/definitions/currency.js](../../src/js/definitions/currency.js)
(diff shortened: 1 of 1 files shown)


docs/building-a-form/common-definitions.md
```diff
@@ -65,37 +65,59 @@ A common type-ahead widget that lets a user type in values and narrow down a lon
...
 
-The common electronic funds transfer (EFT) account information field that collects account type (checking or savings), bank account number, and bank routing number.
+The common electronic funds transfer (EFT) account information field that collects account type (checking or savings), bank account number, and bank routing number. Use these definitions as the uiSchema for the field that represents the account number.
 
-- File: [/src/js/definitions/bankAccount.js](../../src/js/definitions/bankAccount.js)
-- `uiSchema`: Yes
-- `schema`: No
-- Function or object: Object
+```js
+import { uiSchema as bankAccountUI }
```
annekainicUSDS: This has tripped me up before, but I'm pretty sure using the curly braces in the import statement here will cause an issue because there's a default export in `bankAccount.js`. I think instead the import statement should be `import bankAccountUI from 'us-forms-system/lib/js/definitions/bankAccount';`.

docs/building-a-form/common-definitions.md
```diff
@@ -65,37 +65,59 @@ A common type-ahead widget that lets a user type in values and narrow down a lon
...
 ### Currency
 
-- File: [/src/js/definitions/currency.js](../../src/js/definitions/currency.js)
-- `uiSchema`: Yes
-- `schema`: No
-- Function or object: Function
+Formats and validates a US currency field. The display includes a leading `$` character. Call this exported function and pass it the label to be used on the field.
+
+```js
+import { uiSchema as currencyUI }
```
annekainicUSDS: I think this one, as well as most of the definitions, also have default exports, so might want to double check all the import statements in these code examples.

bernars-usa:
This looks great! Very descriptive and helpful. 👍

docs/building-a-form/common-definitions.md
```diff
@@ -65,91 +65,121 @@ A common type-ahead widget that lets a user type in values and narrow down a lon
...
-- `uiSchema`: Yes
-- `schema`: No
-- Function or object: Function
+uiSchema: {
+  serviceStart: monthYearRangeUI()
+}
+```
+Source: [/src/js/definitions/monthYearRange.js](../../src/js/definitions/monthYearRange.js)
 
 ### Non-required full name
```
dmethvin-gov: This one is supposed to be a `schema` modifier but we don't have the actual schema for `fullName` here do we? The starter app has it copied there rather than importing it from USFS. In any case I wonder if we should delete this for now, it's not that important and could be distracting.

docs/building-a-form/common-definitions.md
```diff
@@ -65,91 +65,121 @@ A common type-ahead widget that lets a user type in values and narrow down a lon
...
+Source: [/src/js/definitions/dateRange.js](../../src/js/definitions/dateRange.js)
 
 ### File upload
 
-The file upload field. Requires a specified endpoint for the upload.
+*Note: This is currently not functional. The file upload field requires a specified endpoint for the upload. Documentation will be updated later.*
 
-- File: [/src/js/definitions/file.js](../../src/js/definitions/file.js)
-- `uiSchema`: Yes
-- `schema`: Yes (the same as in vets-json-schema)
```
dmethvin-gov: There's a `schema` in file.js but it looks like nobody uses it and it's just there as a "serving suggestion".

dmethvin-gov: Also, file.js is strange because it does an `export default` on the uiSchema but just plain `export` on the schema. So to import both parts you'd have to do something like:
```js
import { default as fileUI, fileSchema } from '/path/to/file';
```

dmethvin-gov:
I still have address, autosuggest, and personID to document. Should we just drop personID? I saved address and autosuggest for last because they are complicated.

dmethvin-gov:
Just wrestled autosuggest to the ground, at least for a "simple" case. I need to find something already working for address, it's just too complicated to experiment with and I'm also concerned that it isn't the best thing to open with on that page. @annekainicUSDS is there a simple form somewhere that uses it?

Other than address I think this is ready, so feel free to provide feedback now. Even if we can't get address straightened out in time it would be useful to merge this tomorrow before the hackathon.

docs/building-a-form/common-definitions.md
annekainicUSDS: Right, the schema for `fullName` doesn't exist in USFS.

PR #195, opened 2018-07-27
Don't show back button on first page

### LINKED ISSUE #177: Forms without an `introduction` page can never go back
Now that we can have forms without an introduction (see https://github.com/usds/us-forms-system-starter-app/issues/14) it raises the problem of there being a *Back* button on the first page. That button should either be removed or disabled on the first page. The button is [here](https://github.com/usds/us-forms-system/blob/d01100fe70258f42a526e2aa907813e7c4802cc9/src/js/containers/FormPage.jsx#L118-L122) in FormPage.jsx and the `goBack` code that needs to change is [here](https://github.com/usds/us-forms-system/blob/d01100fe70258f42a526e2aa907813e7c4802cc9/src/js/containers/FormPage.jsx#L69-L74).

dmethvin-gov:
Fixes #177

This fix removes the "Back" button from the first page of a form when there is no Introduction component. At the moment, the app crashes when this occurs. 

~~I'm having a hard time getting a good test for this because I don't understand the test setup. When I log out the information going into the FormPage component it seems wrong. The `route.path` is `"testing"` for example but the `location.pathname` is `"/testing"` so those two strings don't match.  When I look at the starter app the pages in the `pageList` seem to have a leading `/` on them.~~

~~Most of the tests seem to be set up with `pathname` of `"/testing/0"` which I also don't understand. That would seem to apply only to array pages? Shouldn't the first page be `"/testing"`?~~

@annekainicUSDS I hope you can help!


annekainicUSDS:
@dmethvin-gov do you also have a sample form config you've been using to test this that I could use as well?

dmethvin-gov:
@annekainicUSDS I don't, I figured I should get the unit test working first. It's possible this is just a quirk of the unit tests but I definitely want to have a test for this. The starter app would work as a test case as well if you delete the `introduction` property.

annekainicUSDS:
@dmethvin-gov I guess I'm wondering how you're testing it? By just writing unit tests, or running a form locally to see the changes? That's what I'm trying to do now, and because the changes haven't been built yet it's not applying to the new changes to the form. For debugging I find it most helpful to run the app locally and place breakpoints in the code to see what's going on as opposed to writing unit tests. Do you mind pushing a commit with the built js files so I can do that? Thanks!

dmethvin-gov:
Sure, I've pushed the built files. Do you usually create unit tests? I think the problem is with the unit test setup but I have no idea what it's doing or why.

test/js/containers/FormPage.unit.spec.jsx
```diff
@@ -196,6 +196,51 @@ describe('Schemaform <FormPage>', () => {
...
+        currentlyLoggedIn: false
+      }
+    };
+
+    const tree = SkinDeep.shallowRender(
+      <FormPage
+        form={form}
+        user={user}
+        route={route}
+        location={location}/>
```
annekainicUSDS: I think the reason why you're seeing the current location path not match the route is because the `location` here is set at the top of this file to be `testing/0`

annekainicUSDS: If you change this to be:
```js
...
location={{pathname: 'testing'}}/>
...
```
the test passes

dmethvin-gov: Yeah, at the top of the file it's set as `location: { pathname: '/testing/0' } }`. I think this is supposed to match the browser's `window.location.pathname` property which is what the router is looking at. Specifically it should have the leading `/` which it does. The problem is that the `route.pageList.path` *does not* have the leading slash so they don't equal. 

Also it has that extra `/0`. I don't know why it is there. It doesn't seem to make any of the tests fail if it is removed, so I can certainly remove that.

I think this may just be a question of getting the router into the right state. Looking at the starter app, it has a [urlPrefix](https://github.com/usds/us-forms-system-starter-app/blob/master/js/config/form.js#L10) that just happens to be `/` by default, so it should work there. It just seems odd that the `path` and `pathname` don't match here. Although I can set the data with a slash-less `location` to make this pass, I still don't understand why and it feels like cheating.

Also, I don't think it's necessary to create the built files when running against the unit tests. The `npm run test` builds everything using Babel before starting the tests. That's super handy and saves lots of time!

dmethvin-gov: It looks like I need to use `formConfig.urlPrefix` in the test, which I've done. If none is set I have it using `'/'` for a default, the test fails if I use `''` and I'm not really sure why. The "it renders" test above expected both the Back and Continue button so I had to change it.

src/js/containers/FormPage.jsx
```diff
@@ -99,6 +99,10 @@ class FormPage extends React.Component {
       // And the data should be for just the item in the array
       data = _.get([route.pageConfig.arrayPath, params.index], data);
     }
+    // It should be "safe" to check that this is the first page because it is
+    // always eligible and enabled, no need to call getPreviousPagePath.
+    const isFirstRoutePage =
+        (route.pageConfig.urlPrefix || '/') + route.pageList[0].path === this.props.location.pathname;
```
annekainicUSDS: I think you shouldn't need to add the `urlPrefix` to the `pageList` path; the code [here](https://github.com/usds/us-forms-system/blob/master/src/js/helpers.js#L78) already does that when generating the `pageList` object to add to the `route` object. When I run this change locally using the starter-app, it's not working as expected because of that.

test/js/containers/FormPage.unit.spec.jsx
annekainicUSDS: To answer a couple of the points of confusion above:

#### When to use '/' or not
In regards to when the prefixed `/` is included or not, I think what happens is the paths are generated from what is specified in the `formConfig`. In the `formConfig`, paths are not written with the prefixed `/`. But in generating the routes, a new object called `pageList` is created, which takes the paths from `formConfig` and adds the `urlPrefix` to them, which in most cases is just a `/`. The router then uses the `pageList` object to determine which is the next and previous pages to go to, and matches that against the `location.pathname`, which always includes the `/` for that path.

For this reason, I think the test code is a bit confusing because we're handwriting the `pageList` object, which is actually generated by the library code. For some reason, the existing test code in this file doesn't include the `/` for the paths in `pageList`, but the actual code would, assuming `urlPrefix` is passed to the `formConfig`. For that reason I think it's a mistake to manually add `urlPrefix` to the path in `pageList` when checking against `location.pathname`, because that should already be included.

#### About the `/0` in the original `location.pathname`
I _think_ this was included because some of the tests are testing array functionality, in which cases items in the array are indexed, and the route tells us which item we're currently on. It's strange that removing this doesn't break the tests, which I think tells me that the test code isn't written accurately in regards to paths. But I need to investigate array paths more...

src/js/containers/FormPage.jsx
dmethvin-gov: Ah, thanks! So if that's the case it sounds like the unit tests are not set up correctly as you mentioned below. I didn't mean to get you into this so deeply but had hoped you might know some background about how this works that would save time. If you want I can take this back and dig into it.

annekainicUSDS: No worries at all! I needed to understand this better myself, so I learned a lot by digging in. Let me know how I can be helpful; if you want to run with making further updates to this that's great, and I can pitch in where needed!

dmethvin-gov: *(grabs baton)*

dmethvin-gov:
@annekainicUSDS This is ready for review now. I changed the mocked routes in the unit tests to have leading slashes the way they would be in the app. I checked and it's working in the starter app as well. I also removed the remnants of the `user` handling since we don't support that. There still is quite a bit of redundancy in these tests (most use the same `form` and `route` setup), it might be worth a pass to clean those up so the tests are easier to read.

PR #212, opened 2018-08-08
Update "Example form.js" document

erikphansen:
## Description

I'm getting up to speed with the US Forms System for my work on Vets.gov with Ad Hoc. I found a small number of errors and areas where clarity was lacking in [the documentation for the form.js config object](https://github.com/usds/us-forms-system/blob/master/docs/building-a-form/quick-start-example-formjs-file.md). This PR addresses them.

## Additional information

I also included a change that reflects an improved/safer naming strategy for both field keys and page ID keys on the config object. It's a change that makes large config objects easier for a human to parse makes them less error prone. Using the same string all over the place is not very safe and leads to errors. Also, using array-access notation to reference the strings that are stored in an object makes it clear that the repeated keys actually are related and are not named the same coincidentally.

### DIFF
--- a/docs/building-a-form/quick-start-example-formjs-file.md
+++ b/docs/building-a-form/quick-start-example-formjs-file.md
@@ -5,9 +5,22 @@
 Use this example `form.js` file to build a basic form. For more information about `form.js`, see "[Creating a form config file](creating-a-form-config-file.md)."
 
 ```js
-{
+// Because pages must be unique across all chapters, it's safer to store them in an object
+const formPages = {
+  pageOne: 'pageOne'
+}
+
+// Since field IDs are repeated in the form config and must be unique throughout the form,
+// it's safer and more convenient to store them in an object
+const formFields = {
+  fieldOne: 'fieldOne',
+  viewFieldTwo: 'view:fieldTwo',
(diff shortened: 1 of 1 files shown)


bernars-usa:
Thanks for working on this, @erikphansen! I've added @annekainicUSDS and @dmethvin-gov as reviewers here.

annekainicUSDS:
@erikphansen Thank you so much for submitting this PR! After reviewing with our team, we're not ready to merge these changes into the library. We think the suggestions for how to structure the form config are potentially very useful to users of the library, but we're not sure these suggestions belong as instructions for how to do things. Instead, we think this suggestion is more appropriate to demonstrate in example forms to show how the library can be used, which we're planning on building out in the next month. For now, I will create [a ticket](https://github.com/usds/us-forms-system/issues/220) to revisit this idea, and close this PR in the meantime. Let us know if you have any additional thoughts or questions on this!

erikphansen:
@annekainicUSDS Totally understandable why you wouldn't want to merge the PR as I wrote it. I'd like to make a much smaller PR that simply fixes a few errors in the example code without injecting my own opinion on how things *should* be done, which rightly doesn't belong in this repo.

PR #215, opened 2018-08-10
Extract definitions 1 built

annekainicUSDS:
## Description

Runs the build on [this PR](https://github.com/usds/us-forms-system/pull/214).


dmethvin-gov:
The hooks are gone from package.json for some reason, although they were committed in https://github.com/usds/us-forms-system/commit/0d79dc9142da5d8225b88df066b7ef177cbcf8d5 . GitHub is returning HTTP 406 when I try to look at the history of the package.json file, I am guessing it got wiped out in some merge conflict.

However (or maybe in addition?) I realize that the pre-push hook doesn't really solve the problem we're trying to solve. If I submit a PR that contains build artifacts but the PR lands after several others in master, the build artifacts in the PR are out of date and missing any PRs landed in the meantime! This would be most worrisome in the case of a release, we might ship stale artifacts.

I've never worked in a repo that committed build artifacts on every commit to master, so I haven't encountered a situation like this before. it seems like we'd need to have the CI re-commit all build artifacts or something, which is starting to get crazy. 

For now I'll just approve this PR but we'll need a fix.

dmethvin-gov:
Hmmm, if we landed PRs by a direct push to master from one of our computers this wouldn't happen because you couldn't land without rebasing against master, and the push hook would require a rebuild at that point. By landing in the GitHub UI we are skipping the push hook.

annekainicUSDS:
Looks like things were removed in this PR: https://github.com/usds/us-forms-system/commit/54593733d288230f71d4d5deb56fab562bfd5264#diff-b9cfc7f2cdf78a7f4b91a753d10865a2.

annekainicUSDS:
Going to merge this PR for now, but agree let's have a larger discussion on the best way to ensure things get built reliably!

PR #219, opened 2018-08-13
Add schemas to remaining definitions

annekainicUSDS:
## Description

Proof of concept of how the new definitions could work with the addition of the schemas. I just added the schemas to a few definitions to start with to make sure the approach looks good before proceeding with the others.

I also used these new definitions in a branch on the starter app to show what it would look like now to use them in the config: https://github.com/usds/us-forms-system-starter-app/compare/test-new-schema-additions-to-usfs?expand=1.

Let me know your initial thoughts on this!


src/js/definitions/dateRange.js
```diff
@@ -1,15 +1,28 @@
 import { validateDateRange } from '../validation';
-import dateUI from './date';
+import { dateConfig, schema as ssnSchema } from './date';
 
-export default function uiSchema(from = 'From', to = 'To', rangeError = 'To date must be after From date') {
+const schema = {
+  type: 'object',
+  properties: {
+    from: ssnSchema,
```
dmethvin-gov: `dateConfig.schema` perhaps?

annekainicUSDS: oops, good catch!

src/js/definitions/dateRange.js
```diff
@@ -1,15 +1,28 @@
 import { validateDateRange } from '../validation';
-import dateUI from './date';
+import { dateConfig, schema as ssnSchema } from './date';
```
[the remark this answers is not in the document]
annekainicUSDS: yes, this was a mistake, thanks for catching!

annekainicUSDS:
@dmethvin-gov I'd still like to test all of these definitions in the starter app to make sure things are working properly, but in the meantime I think this is ready for review!

package.json
```diff
@@ -75,8 +75,7 @@
     "sass-lint": "^1.10.2",
     "sinon": "^6.0.0",
     "skin-deep": "^1.0.0",
-    "uswds": "^1.6.4",
-    "vets-json-schema": "git+https://github.com/department-of-veterans-affairs/vets-json-schema.git"
```
dmethvin-gov: yay!

src/js/definitions/ssn.js
```diff
@@ -2,6 +2,11 @@ import { validateSSN } from '../validation';
 import SSNWidget from '../widgets/SSNWidget';
 import SSNReviewWidget from '../review/SSNWidget';
 
+const schema = {
+  type: 'string',
+  pattern: '^[0-9]{9}$'
+};
+
 const uiSchema = {
```
dmethvin-gov: It seems inconsistent to have some `uiSchema`s pass in their labels and some (like this one) do not. Does it make sense to have all of them always be functions and take their labels/titles as the first arg? 

It's also inconsistent that `schema` are objects but `uiSchema` are not, but I can't think of any good reason to make `schema` a function; if you want to define whether something is `required` or not you can't put it here for example.

annekainicUSDS:
@dmethvin-gov oops, sorry meant to address your concern about the inconsistency between functions and objects! Totally agree that the inconsistency is confusing and it would be better to make things consistent. I would love to leave that to a separate PR if that's okay so we can isolate that change, since it might be significant, and this one already is somewhat significant.

dmethvin-gov:
Yeah, I feel the same way about big PRs so let's land this one and do another go-round.

src/js/definitions/ssn.js
annekainicUSDS: Yes I could imagine some `schema` being functions in the future if you want to specify things like required or `minLength` or something like that. Perhaps it makes sense to make everything a function, even if some things don't have configurable inputs, just for consistency?

Issue #222, opened 2018-08-14
Make definition outputs consistent


annekainicUSDS:
Right now definitions output 2 things: `schema` and `uiSchema`. `schema` is usually an object, and `uiSchema` is sometimes an object, but more often is a function. This means that the way they are both imported in to the form config will be inconsistent, which could cause confusion.

Consider making both `schema` and `uiSchema` functions for consistency.

dmethvin-gov:
Sounds like a good idea, as discussed in #219. 👍

dmethvin-gov:
I guess the `schema` function arguments would be merged into the properties already being returned? So something like:
```js
function schema(options) {
  return { type: 'number', ...options };
}

// used like this?
number.schema();
number.schema({ min: 5 });
```
The `required` attribute is generally on the object *containing* the schema properties that go here, for better or worse, so it can't be set here.

PR #225, opened 2018-08-15
Make definitions consistent

annekainicUSDS:
## Description

Turn all `schema` and `uiSchema` in definitions into functions to make their usage consistent. Here's an example of what that usage would look like: https://github.com/usds/us-forms-system-starter-app/compare/test-new-schema-additions-to-usfs?expand=1.

Connected to https://github.com/usds/us-forms-system/issues/222


annekainicUSDS:
Sounds good! Will merge now and we can always edit/add more later.

bernars-usa:
Are there any documentation updates to make based on these changes?

dmethvin-gov:
Yes indeedy! We'll want to rewrite [common-definitions.md](https://github.com/usds/us-forms-system/blob/4d1d6dc161d94fe5edfe465e601cabe9dc4d6146/docs/building-a-form/common-definitions.md) to explain the new way of using the definitions since they return both `schema` and `uiSchema`. The good news is that they're all consistently functions now and we'll just need to explain what the arguments are but nothing else.

PR #216, opened 2018-08-10
Release 1.1.0

annekainicUSDS:
# 1.2.0 Release

Thank you for using the US Forms System library. For questions about this release, [open an issue](https://github.com/usds/us-forms-system/issues/new/choose).

## New Features

- **Improved library documentation.** This includes Getting Started content, conceptual and referential information about building forms, and reorganized content around customizing forms. Outlined in https://github.com/usds/us-forms-system/issues/86, addressed in https://github.com/usds/us-forms-system/pull/90.
- **Moved a sourcemap embedded in `styles.css` into a separate file.** Outlined in https://github.com/usds/us-forms-system/issues/120, addressed in https://github.com/usds/us-forms-system/pull/121.
- **Improved `FormPage` unit tests.** Addressed in https://github.com/usds/us-forms-system/pull/202.

## Resolved Security Issues

- **Updated `node-sass` to a recent version.** This satisfied the `hoek` vulnerability [CVE-2018-3728](https://nvd.nist.gov/vuln/detail/CVE-2018-3728), reported by GitHub. Fixed in https://github.com/usds/us-forms-system/pull/213.

## Bug Fixes

- **Fields that appeared on both active and inactive pages were incorrectly removed from form submission data.** Reported in https://github.com/usds/us-forms-system/issues/140, fixed in https://github.com/usds/us-forms-system/pull/155.
- **CSS styles for fields that are conditionally expanded below other fields using the `expandUnder` property in the `uiSchema` object were missing from the US Forms System library.** Fixed in https://github.com/usds/us-forms-system/pull/192.
- **In forms without an introduction page, exclude the back button on the first page.** After we made the introduction page optional in the US Forms System Starter App (https://github.com/usds/us-forms-system-starter-app/commit/84c7a23a9372c1f90e08a3e7608505ca5e317606), the back button remained on the first page of the form. Reported in https://github.com/usds/us-forms-system/issues/177, addressed in https://github.com/usds/us-forms-system/pull/195.

PR #234, opened 2018-08-22
Merge production back into master for 1.1.0 release

annekainicUSDS:
## Description

After merging in [the latest release to production](https://github.com/usds/us-forms-system/pull/216), I went to merge production back into master to update master with the upgrade in package version.

However, I encountered [a problem with that PR](https://github.com/usds/us-forms-system/compare/production?expand=1) from production into master, where I noticed two things:

1. It was including 2 commits, both the most recent release and the one before it. It should have only been including the most recent release, because the release before it was already merged into master [here](https://github.com/usds/us-forms-system/pull/146). However, I think this previous commit was accidentally erased at some point because I can't find it in the commit history on master. I know we had encountered a few of Dave's commits that were mysteriously erased at one point, and I think this was another one of those that got erased.

2. The PR was also tracking a bunch of "changes" that were actually current on master. I don't really know why it was doing this, but I assume it is related to the first problem.

The issue of merging the latest on production into master is further complicated by the fact that additional commits have made it into master since the latest release was created, so production and master _should_ be different. We just want to update the change in package number on master.

In order to fix this problem, I created a new branch, resolve-merge-conflict-from-latest-production-release, which is branched off the latest from master. I then merged production into this new branch, and resolved the merge conflicts in two ways:
1. Where there was a merge conflict relating to the package version, I resolved it in favor of what's on production (i.e., upgrading from 1.0.1 to 1.1.0)
2. Where there was a merge conflict relating to anything else (i.e., additional commits that are on master and not yet released), I resolved those in favor of what's on master.

This means master should be updated with the changes on production, but the additional commits on master haven't been overwritten. I'm now creating a PR from resolve-merge-conflict-from-latest-production-release into master.

Hopefully this fixes the problem and subsequent releases don't encounter the same issue. I think we should have a retrospective as a team once @dmethvin-gov is back on how this occurred and how we can prevent it from occurring in the future.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "us-forms-system",
-  "version": "1.0.1",
+  "version": "1.1.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "us-forms-system",
-  "version": "1.0.1",
+  "version": "1.1.0",
   "description": "Build React forms with JSON Schema and the U.S. Web Design System",
   "main": "index.js",
   "directories": {


dmethvin-gov:
Sooo... I think the problem is that the changes in `production` are now squashed into a single commit that represents more than 100 individual commits to master. When we try to merge it back into `master` git looks at the diffs and sees a bunch of changes, some of which conflict (at least *en passant*) when individual `master` commits are applied. If `production` had the same commit hashes as `master` it wouldn't make that mistake and could see the progression.

BTW, I think [this is the full comparison](https://github.com/usds/us-forms-system/compare/production...2e9c1f17afe7bebc1418abc095254084c6cb0dcb) between the two branches, the last commit in master was the one for that security fix.

It seems like we should be using merge commits (rather than squashes) for our syncs between these two branches so that we can see which exact commit hashes are being used. In particular, when merging this back to `master` the position in the history should reflect when the release was actually done and not when this PR was rebased from `production` to `master`. 

However, I still think that for most other situations it makes sense to squash all the commits in a PR, otherwise history looks like a mess (as it does right now) and it's easy to miss problems before landing PRs or track them down after the fact. 

I'm not completely sure what the best resolution is for this.  Technically `production` is now public and we shouldn't change it, so I guess we can't fix the squashed commit there. Most likely we should just land this PR as a merge? This is not something I've dealt with, git is one of those tools with 1,000 different ways to use it and GitHub adds its own set of rules on top of that.

annekainicUSDS:
@dmethvin-gov I did some digging around and asked the Vets.gov team, and I think you're right we should use squash commits for all feature merges into master, and merge commits for releases and syncs between production and master. I think this requires changing a setting in the repo to allow merge commits, which I think we disabled. Enabling it will enable us to decide if we want to squash commit or merge commit, which might mean we will accidentally commit the wrong way at the wrong time (lots of little commits on master for each PR is something we want to avoid), but I don't think we can create rules for different scenarios to restrict certain kinds of commits.

Vets.gov has automatic deployments now where releases are created using tags. They don't do a git branching workflow anymore. Do you think we should be using git tags as well? Even if we want to maintain our git branching workflow we can still add tags. Never used it before so not sure how useful it is!

annekainicUSDS:
Also I think you're right, there may not be much we can do to retroactively fix production, but at least we can do the right thing going forward! And I think this PR will clean up some of the confusion, as it uses the same commit that's on prod now.

dmethvin-gov:
Originally I was going to use a git tag approach and have CircleCI push the release from master, but I can't recall what the issue was there. I'll have to revisit that.

Issue #218, opened 2018-08-10
Strategy for committing build artifacts


dmethvin-gov:
For a fleeting moment we had a git pre-push hook in place but [it didn't do what we wanted](https://github.com/usds/us-forms-system/pull/215#issuecomment-412091307) anyway. 

If we are going to commit build artifacts with every commit, we need some way to ensure that they exactly represent the source committed along with them. It's a bit tricky, the CI can't do it because it's too late by then. We need to have each PR rebased against master at the point where it's landed and *then* rebuild the artifacts before committing. I don't see any [GitHub hooks or API](https://github.com/github/platform-samples) to do that. It's doable if one of us lands the PRs manually.

One possibility is to *not* commit built artifacts except on `production`. That wouldn't affect users of the library since they should be using versioned releases that would still include build artifacts. There are cons to that since it makes testing PRs a two-step process (you have to build after pulling).

jcmeloni-usds:
@dmethvin-gov Did we land on a strategy/plan? I've never been a fan of committing any build artifacts, but don't know the backstory here if there's a reason that we started doing it and it needs to stay.

dmethvin-gov:
I completely agree, I'd prefer to not commit any. Right now vets.gov is pulling in builds via a GitHub hash so they needed it in the past, but I think it's safe to get rid of them now. The npm releases will have everything built.

After our discussions with the vets.gov team yesterday I have some thoughts about how to do this, once we've finished Phase 1 I can get this done.

PR #236, opened 2018-08-27
Update docs for definitions

### LINKED ISSUE #231: Update documentation for definitions
Now that we merged [this PR](https://github.com/usds/us-forms-system/pull/225), which changes how the definitions are used, we need to update our documentation.

There were a few main changes to how definitions work that we need to address:

1. For the remaining definitions (I also deleted some), they used to just include `uiSchema`. I added the `schema` as well, so they now export an object with both the `schema` and `uiSchema`.
2. The naming convention I've followed for all definitions is each one exports an object which is the name of the definition + `Config`. Example: the currency definition exports an object called `currencyConfig`. You can verify what each exported object is called by looking at the bottom of each definition file. The exported object returns 2 things: the `schema` and `uiSchema`, which are named the same for each definition.
3. I made all `schema` and `uiSchema` functions, whereas before some were functions and some were just objects. This means that the way you use them is slightly different. When they were just objects, you would use them in the config by just listing the name (e.g., for the ssn field (https://github.com/usds/us-forms-system/blob/master/docs/building-a-form/common-definitions.md#social-security-number), after importing the `uiSchema` at the top of the file, you would call it in the config by writing `ssnUI`. Now, you need to call it by first referring to the exported object (`ssnConfig`), calling the schema or uiSchema within it using dot notation (`ssnConfig.uiSchema`), and then adding parens to the end (`ssnConfig.uiSchema()`).
4. Some `uiSchema` take optional arguments in the function (e.g., `ssnConfig.uiSchema` optionally takes a title, so if you wanted to change the title from the default you would pass your title to the function like so: `ssnConfig.uiSchema('New title')`). Not all `schema` and `uiSchema` take optional arguments though. You can see which ones do and what their arguments are by referring to the definition files and seeing which functions include an argument in the parens for `schema` and `uiSchema`.

For the new documentation, I don't think you need to include the level of detail I included above; that was mainly to make sure it's clear to who will be documenting this! I think just describing that each definition exports an object that includes both `schema` and `uiSchema`, and that both `schema` and `uiSchema` are functions will be sufficient. Then showing with a code sample how you call those `schema` and `uiSchema` for each definition will be useful, and including info on which ones take optional arguments.

For an example of how they're all called, you can look at this test branch on the starter-app: https://github.com/usds/us-forms-system-starter-app/compare/test-new-schema-additions-to-usfs?expand=1.

Let me know if there are any questions!

bernars-usa:
Closes #231, which is related to #225.

This PR:
- Updates the conceptual information about definitions
- Updates the individual descriptions of the definitions that were changed in #225 (everything except Autocomplete and File Upload)
- Updates the code sample for how to call the definitions based on the sample `form.js` in https://github.com/usds/us-forms-system-starter-app/blob/test-new-schema-additions-to-usfs/js/config/form.js
- Adds images of each field to show how they're rendered

**This needs a thorough review of the code samples, as well as the descriptions of the definitions.**


docs/building-a-form/common-definitions.md
```diff
@@ -2,27 +2,29 @@
...
-- More complex definitions are functions that require certain parameters.
+Definitions export `schema` and `uiSchema` objects, which are functions that require certain parameters.
 
 ### Using definitions
 
 To use a definition, import it near the top of the file you want to reference it in:
 
 ```js
-import currencyUI from 'us-forms-system/lib/js/definitions/currency';
+import currencyConfig from 'us-forms-system/lib/js/definitions/currency';
```
annekainicUSDS: This needs to be imported using curly brackets now: `import { currencyConfig } from ...`. The reason is because before we were using default exports in the definitions file, and now we're not (basically changing how we export/import things).

docs/building-a-form/common-definitions.md
```diff
@@ -2,27 +2,29 @@
...
 ```
 
-Then, call it to add all the `uiSchema` definitions. In this example, the definition is a function that takes the title for that field, which is used to populate the 'ui:title' property in uiSchema:
+Then, call it to add the `schema` and `uiSchema` functions. The definition is a function that takes the title for that field, which is used to populate the 'ui:title' property in uiSchema:
```
annekainicUSDS: I think we need to change this language a little bit. It's still pretty specific to `uiSchema`. Here's my recommended edit:

Then, use the definition by calling the `schema` and `uiSchema` functions on it where needed. Sometimes the functions take arguments to configure how the definitions are used. In this particular case, the `uiSchema` for currency takes an optional argument for the title of the field. In this example we want to set the title to "Currency", so we pass that as a string to the `uiSchema` function.

bernars-usa: 🤔 I can't remember making this change... it might have been a half implementation of another thought. Let me know if it needs changing.

docs/building-a-form/common-definitions.md
```diff
@@ -45,22 +47,25 @@ Available definitions are:
 ### Autosuggest
 
 A common type-ahead widget that lets a user type in values and narrow down a longer list of options. It is most commonly used with an `enum` of the available options as shown here. Define the uiSchema by calling the function that you import. You can pass an object with additional uiSchema properties.
+
 ```js
 import { uiSchema as autosuggestUI } from 'us-forms-system/lib/js/definitions/autosuggest';
```
annekainicUSDS: This import statement needs to change to `import { autosuggestConfig } from 'us-forms-system/lib/js/definitions/autosuggest';`.

docs/building-a-form/common-definitions.md
```diff
@@ -45,22 +47,25 @@ Available definitions are:
...
-      type: 'string',
-      enum: [
-        'LA', 'NY', 'CH'
-      ],
-      enumNames: [
-        'Los Angeles',
-        'New York',
-        'Chicago'
-      ]
+function schema() {
```
annekainicUSDS: I think what these code samples are trying to show is what this would look like in the form config, not what the definition code is. So I think this would actually remain the way it was before, where the `schema` is an object in the form config. You can just revert this change for the `schema`.

This particular definition is a little confusing because here we're not actually using the `schema` from the definition, but instead writing our own schema into the form config, so I think we should clean that up as a separate issue. I'll create a ticket for that.

docs/building-a-form/common-definitions.md
```diff
@@ -45,22 +47,25 @@ Available definitions are:
...
+        ],
+        enumNames: [
+          'Los Angeles',
+          'New York',
+          'Chicago'
+        ]
+      }
     }
   }
 },
```
annekainicUSDS: The usage for the `uiSchema` below also needs to change, where instead it would look like this:

```
uiSchema: {
  officeLocation: autosuggestConfig.uiSchema(
    ...
  )
}
```

(what is contained in the `...` is exactly as it is now (lines 74-81), it's just where the uiSchema function is called that needs to change)

docs/building-a-form/common-definitions.md
```diff
@@ -81,63 +86,105 @@ Source: [/src/js/definitions/autosuggest.js](../../src/js/definitions/autosugges
 
 ### Currency
 
-Formats and validates a US currency field. The display includes a leading `$` character. Call this exported function and pass it the label to be used on the field.
+Formats and validates a US currency field that includes a leading `$` character. You can pass this function a label to be used on the field.
```
annekainicUSDS: I would change the language in the second sentence to be: "You can pass the `uiSchema` function a label to be used on the field".

docs/building-a-form/common-definitions.md
```diff
@@ -81,63 +86,105 @@ Source: [/src/js/definitions/autosuggest.js](../../src/js/definitions/autosugges
...
+    currency: currencyConfig.schema()
+  }
 }
 ```
 Source: [/src/js/definitions/currency.js](../../src/js/definitions/currency.js)
 
 ### Current or past dates
 
-The common date field with current or past validation set (i.e., dates in the future are not valid). Call this exported function and pass it the label to be used on the field.
+The common date field with validation warning that dates in the past or future are not valid. You can pass this function a label to be used on the field. Dates must be on or before January 1, 1900.
```
annekainicUSDS: Anywhere where it says "You can pass this function a label", I would specify to be "You can pass the `uiSchema` function a label" because there are now 2 functions that are exported by a definition: the `schema()` and the `uiSchema()`. It's important to specify which of the two takes the additional argument.

docs/building-a-form/common-definitions.md
```diff
@@ -81,63 +86,105 @@ Source: [/src/js/definitions/autosuggest.js](../../src/js/definitions/autosugges
...
+
+![Four date fields, with and without focus, with an invalid year, and with valid data](../images/definitions/currentOrPastDate.png)
 
 ```js
-import currentOrPastDateUI from 'us-forms-system/lib/js/definitions/currentOrPastDate';
+import { currentOrPastDateConfig } from 'us-forms-system/lib/js/definitions/currentOrPastDate';
 
 uiSchema: {
-  birthdate: currentOrPastDate('Date of Birth')
+  currentOrPastDate: currentOrPastDateConfig.uiSchema()
```
annekainicUSDS: Do we want to show how this takes an additional argument for the label like we used to? So:

```
currentOrPastDate: currentOrPastDateConfig.uiSchema('Date of Birth')
```

docs/building-a-form/common-definitions.md
```diff
@@ -81,63 +86,105 @@ Source: [/src/js/definitions/autosuggest.js](../../src/js/definitions/autosugges
...
+
+![Three date fields, one empty, one with invalid data, and one with valid data](../images/definitions/currentOrPastMonthYear.png)
 
 ```js
-import currentOrPastMonthYear from 'us-forms-system/lib/js/definitions/currentOrPastMonthYear';
+import { currentOrPastMonthYearConfig } from 'us-forms-system/lib/js/definitions/currentOrPastMonthYear';
 
 uiSchema: {
-  lastContact: currentOrPastMonthYear('Last Contact')
+  currentOrPastMonthYear: currentOrPastMonthYearConfig.uiSchema()
```
annekainicUSDS: Same thing about showing the label that can be passed to it

docs/building-a-form/common-definitions.md
dmethvin-gov: I'd keep the `birthdate` as well since it's a good example of when you might want a date not in the future (assuming the person filling out the form is giving their own birthdate 🤔).

docs/building-a-form/common-definitions.md
dmethvin-gov: The `lastContact` was similarly trying to give an example of when you might want this type of validated date.

docs/building-a-form/common-definitions.md
bernars-usa: Thank you!

docs/building-a-form/common-definitions.md
annekainicUSDS: Ah yes, good point @dmethvin-gov!

docs/building-a-form/common-definitions.md
```diff
@@ -61,83 +66,119 @@ schema: {
         'New York',
         'Chicago'
       ]
+      }
```
annekainicUSDS: I think this one last } should be deleted!

docs/building-a-form/common-definitions.md
```diff
@@ -61,83 +66,119 @@ schema: {
...
-    'Preferred Office Location',  // field title
-    null,         // Promise to get options (optional)
-    {             // Additional uiSchema options
-      'ui:options': {
-        // When labels are not provided, it uses enumNames
-        labels: { }
-      }
-    }
+  officeLocation: autosuggestConfig.uiSchema(
+    ...
```
annekainicUSDS: Oops, sorry! I think my previous feedback was confusing. This is what the code should look like here:

```
uiSchema: {
  officeLocation: autosuggestConfig.uiSchema(
    'Preferred Office Location',  // field title
    null,         // Promise to get options (optional)
    {             // Additional uiSchema options
      'ui:options': {
        // When labels are not provided, it uses enumNames
        labels: { }
      }
    }
  )
}
```

docs/building-a-form/common-definitions.md
```diff
@@ -61,83 +66,119 @@ schema: {
...
+import { currentOrPastDateConfig } from 'us-forms-system/lib/js/definitions/currentOrPastDate';
 
 uiSchema: {
-  birthdate: currentOrPastDate('Date of Birth')
+  birthdate: currentOrPastDateConfig.uiSchema('Date of Birth')
+},
+schema: {
+  type: 'object'
+  properties: {
+    currentOrPastDate: currentOrPastDateConfig.schema()
```
annekainicUSDS: This key also needs to be `birthdate`, so:

```
        birthdate: currentOrPastDateConfig.schema()
```

docs/building-a-form/common-definitions.md
```diff
@@ -61,83 +66,119 @@ schema: {
...
+import { currentOrPastMonthYearConfig } from 'us-forms-system/lib/js/definitions/currentOrPastMonthYear';
 
 uiSchema: {
-  lastContact: currentOrPastMonthYear('Last Contact')
+  lastContact: currentOrPastMonthYearConfig.uiSchema('Last Contact')
+},
+schema: {
+  type: 'object',
+  properties: {
+    currentOrPastMonthYear: currentOrPastMonthYearConfig.schema()
```
annekainicUSDS: This key also needs to be `lastContact`, the same as above.

docs/building-a-form/common-definitions.md
```diff
@@ -61,83 +66,119 @@ schema: {
...
+import { dateConfig } from 'us-forms-system/lib/js/definitions/date';
 
 uiSchema: {
-  startDate: dateUI('startDate')
+  startDate: dateConfig.uiSchema('Start Date')
+},
+schema: {
+  type: 'object',
+  properties: {
+    date: dateConfig.schema()
```
annekainicUSDS: `startDate` here instead of `date`

docs/building-a-form/common-definitions.md
```diff
@@ -61,83 +66,119 @@ schema: {
...
+import { dateRangeConfig } from 'us-forms-system/lib/js/definitions/dateRange';
 
 uiSchema: {
-  servicePeriod: dateRangeUI('servicePeriod')
+  servicePeriod: dateRangeConfig.uiSchema('Period of Service')
+},
+schema: {
+  type: 'object',
+  properties: {
+    dateRange: dateRangeConfig.schema()
```
annekainicUSDS: `servicePeriod` instead of `dateRange`

docs/building-a-form/common-definitions.md
```diff
@@ -150,60 +191,105 @@ Source: [/src/js/definitions/file.js](../../src/js/definitions/file.js)
...
+import { monthYearConfig } from 'us-forms-system/lib/js/definitions/monthYear';
 
 uiSchema: {
-  serviceStart: monthYearUI('Month/Year Service Started')
+  serviceStart: monthYearConfig.uiSchema('Month and Year Service Started')
+},
+schema: {
+  type: 'object',
+  properties: {
+    monthYear: monthYearConfig.schema()
```
annekainicUSDS: `serviceStart` instead of `monthYear`

docs/building-a-form/common-definitions.md
```diff
@@ -150,60 +191,105 @@ Source: [/src/js/definitions/file.js](../../src/js/definitions/file.js)
...
+import { monthYearRangeConfig } from 'us-forms-system/lib/js/definitions/monthYearRange';
 
 uiSchema: {
-  serviceStart: monthYearRangeUI()
+  serviceRange: monthYearRangeConfig.uiSchema('Period of Service')
+},
+schema: {
+  type: 'object',
+  properties: {
+    monthYearRange: monthYearRangeConfig.schema()
```
annekainicUSDS: `serviceRange` instead of `monthYearRange`

docs/building-a-form/common-definitions.md
```diff
@@ -150,60 +191,105 @@ Source: [/src/js/definitions/file.js](../../src/js/definitions/file.js)
...
+import { yearConfig } from 'us-forms-system/lib/js/definitions/year';
 
 uiSchema: {
-  yearBorn: yearUI
+  taxYear: yearConfig.uiSchema('Year Filed')
+},
+schema: {
+  type: 'object',
+  properties: {
+    year: yearConfig.schema()
```
annekainicUSDS: `taxYear` instead of `year`

docs/building-a-form/common-definitions.md
bernars-usa: Many thanks!

docs/building-a-form/common-definitions.md
bernars-usa: It didn't look right but I couldn't figure out how to check it. Thanks!

bernars-usa:
Thanks for the reviews, @annekainicUSDS and @dmethvin-gov! This is ready for a final check.

dmethvin-gov:
it was close to working and I managed to mess it up while resolving this spurious conflict with the map file. This is a result of #218 where our build artifacts don't track the real source files. I'll mess with it and try to get something I can commit.

dmethvin-gov:
Not sure why this failed in CircleCI the first time but I re-ran it and it's okay.

Issue #211, opened 2018-08-08
Eliminate vets.gov-isms in form submit and file upload


dmethvin-gov:
There are still quite a few things that depend on server behavior specific to vets.gov like the use of ` req.setRequestHeader('X-Key-Inflection', 'camel')` [in actions.js](https://github.com/usds/us-forms-system/blob/2410134ffda3767058ae7ef16e512e3fc5124632/src/js/actions.js#L124) that expects the use of [particular Rails middleware](https://github.com/department-of-veterans-affairs/vets-website/issues/2953). Since servers and their configurations are so different this might be a good place to just have the app pass us a callback in `formConfig`. We pass it the `formData` and they use to submit the form, serializing it whatever way they want (JSON, XML, form-urlencoded). The same goes for file uploads.

dmethvin-gov:
Adding the `va-growable-*` classes in [m-schemaform.scss](https://github.com/usds/us-forms-system/blob/master/src/scss/modules/_m-schemaform.scss). I suppose these might just need to be renamed to be generic but they may be defining some style we want to modify.

jcmeloni-usds:
@dmethvin-gov Confirming this is still ongoing -- correct?

dmethvin-gov:
Yes, I'll be making some changes once the new router lands.

jcmeloni-usds:
Ok, moved to phase 2 and the 2.0.0-alpha1 milestone

annekainicUSDS:
Another thing to add here: the "Submit again" button that only appears in dev and staging, we want to remove that now.

dmethvin-gov:
Adding the rework of `stringifyFormReplacer` from #259, which currently has special case processing for various widgets.

PR #242, opened 2018-08-31
[WIP] Followup to review comments on #209

dmethvin-gov:
* Add more docs (/cc @bernars-usa)
* Reduce drilled props from 100 to 99 or so
* Centralize defaults for preSubmitInfo to simplify logic
* Reduce magic in SubmitController tests by switching to enzyme
  - This requires deep rendering on several tests but seems more realistic than reaching into the instance and calling lifecycle or event methods!
* Insert some comments with concerns and observations

@annekainicUSDS I was confused by a few things and marked them in BUG or TODO comments. You may see a few comments from me on this PR as I try to figure them out.


docs/building-a-form/available-form-features-and-usage-guidelines.md
```diff
@@ -419,14 +419,29 @@ For the code implementation, see the [`review` folder](../../src/js/review).
 
 ### Required checkbox before form submission
 
-Use this feature to require a user to indicate they have read terms and conditions, a privacy policy, or any other text before submitting your form. It includes a checkbox and short-form text that can include relevant links to more verbose information on separate pages on your site. To configure this feature, place a `preSubmitInfo` object in the `formConfig`:
+Use this feature to require a user to indicate they have read terms and conditions, a privacy policy, or any other text before submitting your form. It includes a checkbox and short-form text that can include relevant links to more verbose information on separate pages on your site.
+
+To configure this feature, place a `preSubmitInfo` object in the `formConfig`. These are the available options:
+<dl>
```
bernars-usa: This is great! In recent years, I've leaned away from `<dl></dl>` in Markdown because it's a lot more difficult to maintain (opting for plain list items with strong emphasis on the term instead). I'll leave that up to you, though.

docs/building-a-form/available-form-features-and-usage-guidelines.md
```diff
@@ -419,14 +419,29 @@ For the code implementation, see the [`review` folder](../../src/js/review).
...
+<dt><code>notice</code></dt>
+<dd>A text string or <a href="https://reactjs.org/docs/rendering-elements.html">React element</a> that is placed above the checkbox and submit button. If the form definition file is `.jsx` the definition can be inline, or use `import` to reference an external component. Optional; if nothing is given for this property no notice appears.</dd>
```
bernars-usa: > Optional; if nothing is given for this property no notice appears.

It's a good idea to state whether the properties are optional up front, like so:

"`notice` (Optional): A text string or React element that..."

docs/building-a-form/available-form-features-and-usage-guidelines.md
```diff
@@ -435,8 +450,8 @@ preSubmitInfo: {
 
 #### Usage guidelines
 
-Right now, the required checkbox is automatically included in all forms. The US Forms System team will refactor this component to make it more customizable. To follow that discussion, subscribe to https://github.com/usds/us-forms-system/issues/53.
+No notice or checkbox appears above the submit button unless a `preSubmitInfo` section is specified. Most applications will want, at minimum, to give some sort of notice to the user before the form is submitted. Although this section is optional, it is highly recommended.
```
bernars-usa: > No notice or checkbox appears above the submit button unless a `preSubmitInfo` section is specified. Most applications will want, at minimum, to give some sort of notice to the user before the form is submitted. Although this section is optional, it is highly recommended.

Less passive:

"Unless you specify a `preSubmitInfo` section, no notice or checkbox appears above the submit button. Most applications will want, at minimum, to give some sort of notice to the user before they submit the form. Although this section is optional, we recommend you specify it."

src/js/review/SubmitController.jsx
```diff
@@ -51,35 +64,36 @@ class SubmitController extends React.Component {
...
-    let errors;
+    const preSubmit = this.getPreSubmit(formConfig);
+    let isValid = true;
 
-    // If a pre-submit agreement was specified, it has to be accepted first
-    const preSubmitField = formConfig.preSubmitInfo &&
-        formConfig.preSubmitInfo.required && (formConfig.preSubmitInfo.field || 'AGREED');
-    if (preSubmitField && !form.data[preSubmitField]) {
+    // If a pre-submit agreement is required, make sure it was accepted
+    if (preSubmit.required && !form.data[preSubmit.field]) {
```
annekainicUSDS: This if/else block is a bit confusing to me. The logic I see is the following:
```
if preSubmit is required AND is not true in the form data
  isValid = false
else (it's either not required OR (it is required AND it is true in the form data))
  run the isValidForm check and set isValid to whatever the output is
  
  if isValid is false
    record and track information about it

if isValid is true
  submit the form
else (isValid is false)
  set an attempted submit status
```
So it seems like there are 2 main if/else blocks here:
1. The first is used to set the value of `isValid`
2. The second is used to do things based on the value of `isValid`

What I think is a bit confusing is there's logic within the first block that does things based on the value of `isValid` (tracking events). Would it be clearer if that was contained in the second block?

src/js/review/SubmitController.jsx
```diff
@@ -89,21 +103,21 @@ class SubmitController extends React.Component {
...
+    const preSubmit = this.getPreSubmit(formConfig);
+
     return (
       <div>
-        { this.preSubmitInfo && <PreSubmitSection
-          required
-          preSubmitInfo={formConfig.preSubmitInfo}
-          onChange={() => this.props.setPreSubmit(formConfig.preSubmitInfo.field, this.value)}
-          form={form}
+        <PreSubmitSection
```
annekainicUSDS: So now that we're always initializing `preSubmit` to _something_ through the `getPreSubmit` thing at the top of this file, we don't need to check whether or not to render this component based on what's in the `formConfig`? Because we initialize it to `false`, that will mean if there's nothing else specified in the `formConfig` nothing will actually be rendered for the checkbox?

dmethvin-gov: Right, if nothing is specified we'll have a `notice` that is empty (so it renders nothing) and a `required` that is not `true` (it's `undefined`) so the checkbox and label won't render.

dmethvin-gov:
Right now I can run through with the debugger and see that it this working properly, but it crashes on submit since there is nowhere to submit it to. We should set up the starter app so it just does something harmless with the data like dump it in `JSON.stringify` format to the page. Probably best to wait until #211 is done before tackling that. (All this stuff is in the same few files so changes end up conflicting!)

Issue #152, opened 2018-07-11
Reduce production file size


dmethvin-gov:
I've added `webpack-bundle-size` to the starter app and it generates output into `public/stats/bundle.html` on each production build showing the sizes of components in the bundle. I can't upload a picture of the web page it generates (GitHub uploads are blocked) so I'll summarize the main issues here. The biggest dependencies by size are:

| Component  |   Parsed size (KB)    | Pulled in via |
|--------------|-------------------:|-------------|
|   Total            |  1001    |  |
| moment        |  227  | us-forms-system |
| styles.css     |  181   | us-forms-system (and USWDS) |
| USFS JS code | 119 | us-forms-system |
| react-dom   | 101 | react-jsonschema-form |
| lodash          |  55  | us-forms-system and others |
| react-jsonschema-form | 53 | us-forms-system |
| raven-js       |    39  |  us-forms-system |
| react-scroll  | 24 | us-forms-system |

So the sorta-good news is that most of these are direct dependencies so we possibly have the ability to shrink or remove them. Here are my notes.

- [x] **moment:** (Fixed in #258) One of moment's benefits is that it knows about internationalization of dates. We're pulling date localizations for every language and country yet our forms seem to be hardwired to English. We're actually using moment for [almost nothing](https://github.com/usds/us-forms-system/search?q=moment) so we should be able to get rid of it with a few custom functions.

- [ ] **styles.css:** This is a *lot* of CSS and it's not that well compressed. If there are parts of it that we know we don't need for forms, it would be awesome to be able to remove them. In the case where us-forms-system is used within a larger app, that app may want to pull in styles that we don't use so it's not as simple as chopping things out. This relates in some ways to #65 and #66.

- [ ] **USFS JS code:** The sample app pulls in almost no widgets or definitions from us-forms-system so it's a bit worrisome that it's already so big. Not sure if there's a way to avoid this growth.

- [ ] **react-dom:** The only function we seem to directly use from this in production is `render()`but I suspect there is a lot used under the covers for JSX parsing and processing. I suppose this is 100KB that we have to swallow.

- [x] **lodash:** I've already mentioned this in detail in #131.

- [ ] **react-jsonschema-form:** If we extract only the parts we need, as recommended in #141, this may save us a decent amount of space. 

- [x] **raven-js:** (Fixed in #196) This is an error reporting library. See #134 for detailed discussions, but I'd prefer to have just some hooks in the library and let the app decide how much tracking they want.

[...]

PR #258, opened 2018-09-09
Use date-fns instead of moment

timwis:
Helps address #152, replacing the `227 KB` of moment.js with `29 KB` of [date-fns](https://date-fns.org/), a popular utility library for date functions that's structured similarly to lodash.

Please note that I don't believe all the affected functions have unit tests, and I'm not deeply familiar with this codebase, so I'd appreciate a second set of eyes on it!

Also note that we may be able to remove `timeFromNow` in favour of date-fn's [`distanceInWordsToNow`](https://date-fns.org/v1.29.0/docs/distanceInWordsToNow) and possibly a few other DIY date functions in here, but I wanted to limit the scope of this PR.

## Types of changes

- [x] Refactor

## Checklist:

- [ ] I added tests to verify a bug fix or new functionality. (N/A)
- [ ] My change requires a change to the documentation, and I've updated the documentation accordingly. (N/A)

screen shot 2018-09-09 at 18 56 06


src/js/actions.js
```diff
@@ -235,7 +234,8 @@ export function uploadFile(file, uiOptions, onProgress, onChange, onError) {
       } else {
         let errorMessage = req.statusText;
         if (req.status === 429) {
-          errorMessage = `You’ve reached the limit for the number of submissions we can accept at this time. Please try again in ${timeFromNow(moment.unix(parseInt(req.getResponseHeader('x-ratelimit-reset'), 10)))}.`;
+          const resetDate = new Date(req.getResponseHeader('x-ratelimit-reset') * 1000);
```
dmethvin-gov: Noting that the date conversion here looks right since the original was in seconds-since-epoch and JS is milliseconds-since-epoch. There [seems to be some controversy](https://github.com/zalando/restful-api-guidelines/issues/379) about whether this is a *relative* time or an *absolute* one. I can see the point about absolute being risky because it assumes relatively synchronized clocks. In any case, this still does whatever it did before, whether right or wrong. 😺

src/js/review/SubmitButtons.jsx
```diff
@@ -62,7 +61,7 @@ export default function SubmitButtons(props) {
       <div className="usa-alert usa-alert-error schemaform-failure-alert">
         <div className="usa-alert-body">
           <p className="schemaform-warning-header"><strong>We’ve run into a problem</strong></p>
-          <p>We’re sorry. Your submission didn’t go through because we received too many requests from you. Please wait {timeFromNow(moment.unix(submission.extra))} and submit your request again.</p>
+          <p>We’re sorry. Your submission didn’t go through because we received too many requests from you. Please wait {timeFromNow(new Date(submission.extra * 1000))} and submit your request again.</p>
```
dmethvin-gov: Weird, this is basically the same message but phrased a different way?

src/js/utilities/date/index.js
```diff
@@ -38,46 +43,38 @@ function formatDiff(diff, desc) {
...
  * the provided date occurs. It’s meant to be less fuzzy than moment’s
  * timeFromNow so it can be used for expiration dates
  *
- * @param date {Moment Date} The future date to check against
- * @param userFromDate {Moment Date} The earlier date in the range. Defaults to today.
+ * @param date {Date} The future date to check against
+ * @param userFromDate {Date} The earlier date in the range. Defaults to today.
  * @returns {string} The string description of how long until date occurs
  */
 export function timeFromNow(date, userFromDate = null) {
```
dmethvin-gov: Just noting for future optimization that the only place where we use `timeFromNow` (and thus all the `differenceIn*` from `date-fns` is the "Please don't push Submit so often" message. Subtracting the two dates and expressing in terms of integer minutes is probably all we need there. I think we should land this as-is though and work on these details later.

timwis:
@dmethvin-gov no worries, should be all sorted now. :+1:

Issue #266, opened 2018-09-14
On the review page, pagePerItem pages with `depends` functions are not correctly passed `index`


jbalboni:
**Describe the bug**
Normally, the `depends` property in a page config is passed `index` as a second parameter for `pagePerItem` pages. But in `ReviewChapters`, we run the `depends` logic before those `index` properties are set, because we call `getActivePages` before `expandArrayPages`. In `routing.js` we do the opposite, so the logic works correctly outside of the review page.

The result is that pages that have a `depends` function that needs `index` will be shown/hidden incorrectly on the review page.

**To Reproduce**
Use a page config with the following options:

```
{
  path: 'some-url/:index`,
  showPagePerItem: true,
  arrayPath: 'someArray',
  depends: (formData, index) => typeof index !== 'undefined'
}
```
Pages will be visible on normal form pages, but not on the review page.

**Expected behavior**
`index` is populated when calling `depends`

jcmeloni-usds:
@jbalboni Confirming that PR #267 addressed this and we can close this? (thanks, if so!)

jbalboni:
@jcmeloni-usds Yep, this can be closed.

jcmeloni-usds:
Yay! Thanks.

PR #267, opened 2018-09-14
Fix expand active order in ReviewChapters

jbalboni:
This swaps the order of `getActivePages` and `expandArrayPages` to match `routing.js`.

<Addresses/Closes> #266

## Types of changes

## Checklist:


jbalboni:
Will push up a test and prod files shortly.

src/js/review/ReviewChapters.jsx
```diff
@@ -125,8 +125,8 @@ function mapStateToProps(state, ownProps) {
   const chapters = chapterNames.map(chapterName => {
     const pages = pagesByChapter[chapterName];
 
-    const activePages = getActivePages(pages, formData);
-    const expandedPages = expandArrayPages(activePages, formData);
+    let expandedPages = expandArrayPages(pages, formData);
```
dmethvin-gov: Why `let` here and not `const`? You can always nest the `expandArrayPages()` in the next line.

jbalboni: Nesting seemed less readable than this, but I'm happy to change it if you'd prefer.

jbalboni:
@dmethvin-gov Yeah, looks like those need changing and also the helper function seems like a good idea.

src/js/review/ReviewChapters.jsx
```diff
@@ -106,7 +104,7 @@ class ReviewChapters extends React.Component {
   }
 }
 
-function mapStateToProps(state, ownProps) {
+export function mapStateToProps(state, ownProps) {
```
jbalboni: Not sure how you all feel about this. We made the changes in here to put more code in mapStateToProps because it's more testable, but in order to test it directly, we need to export it.

jbalboni:
@dmethvin-gov Should be ready for another look.

src/js/review/ReviewChapters.jsx
dmethvin-gov: More testing is always good. In this case it's testing a lot of implementation and plumbing details so not ideal, but the test is clearly marked so if it breaks during a refactor it should be clear why.

dmethvin-gov:
This duo of calls seems to be everywhere, [here's another one in FormNav](https://github.com/usds/us-forms-system/blob/ffd15c43556ab4108c1481b8d6163c8a1589a1ea/src/js/components/FormNav.jsx#L27-L30). Can you look around and see if I missed any others? Would it be possible to not export one or both of the two individual functions, are they used separately for any reason?

jbalboni:
I found the FormNav one when I took another pass through, but didn't see any others. I did see some uses of the functions on their own, but I didn't look too deeply at if that was valid. I will take another look at that on Monday.

jbalboni:
@dmethvin-gov I looked through for uses of just `getActivePages` and it looks like the default `transformForSubmit` was using it without expanding array pages first. So I've updated that, too.

I think all this code was missing because `itemFilter` instead of `depends` was initially meant for filtering out items in `pagePerItem` pages, but then the index was added to `depends` for one of our forms and wasn't full fleshed out. On vets we don't have any uses of `depends` and `pagePerItem` pages outside of one form, which is only now going to beta.

src/js/helpers.js
```diff
@@ -578,3 +580,17 @@ export function omitRequired(schema) {
 
   return newSchema;
 }
+
+/*
+ * Normal transform for schemaform data
+ */
+export function transformForSubmit(formConfig, form, replacer = stringifyFormReplacer) {
+  const expandedPages = expandArrayPages(createFormPageList(formConfig), form.data);
+  const activePages = getActivePages(expandedPages, form.data);
```
annekainicUSDS: Why are we not using `getActiveExpandedPages` here if it's doing these same 2 checks in this order? Because we need to calculate `expandedPages` separately for the next function?

jbalboni: Yeah, I thought it was useful to used the expandedPage array for both of the following function calls and make it clear that we're using the same base data to get the active and inactive page lists.

annekainicUSDS: Yep, that's what it seemed like. Feel free to merge this when you're ready @jbalboni!

jbalboni:
@annekainicUSDS I think you'll have to merge it, I don't have write access here.

annekainicUSDS:
@jbalboni Oops sorry! Will merge now.

Issue #252, opened 2018-09-05
Create exemplar form: DS-82


dmethvin-gov:
See discussion in #204.

U.S. Passport Renewal Application for Eligible Individuals (https://eforms.state.gov/Forms/ds82.pdf) 

This one also has an online "form filler" that allows the user to fill out a digital form, then generates a PDF to print and mail in. (https://pptform.state.gov/?Submit2=Complete+Online+%26+Print)

dmethvin-gov:
Like #250, this form has a worksheet in the instructions to help people understand whether they should be filling out this form in the first place. It seems like a trend may be forming here?

This is also a form that probably can't be completed online as well since it requires the applicant to attach a photo.

dmethvin-gov:
Here's the form so far: https://gist.github.com/dmethvin-gov/41ae189dd88213e5e89f81452a0d2a23

Things that need to be resolved:

* The first (qualifications) page doesn't show an error if you fail to check boxes before moving to the next page, but it *does* prevent you from moving to the next page. I guess that's why there is an ErrorableCheckbox? It seems odd that a checkbox could fail the `required` validation but not show an error. I can probably get around this by switching to YesNoWidget but thought that checkboxes looked better.
* We could definitely use some canned standard address field definitions, especially for US states.
* What is our story for forms like this that definitely seem to require a paper output? It seems like it would be super useful to have an (optional!) component that supported printed forms along with a printed checklist of "here's what you need to do to complete the form." Several of these forms we've chosen as exemplars are in this category.

Issue #251, opened 2018-09-05
Create exemplar form: SS-5


dmethvin-gov:
See discussion in #204.

Application for a Social Security Card (https://www.ssa.gov/forms/ss-5.pdf)

dmethvin-gov:
Wow, this is a surprisingly complex form. I'll just note several things here that I've puzzled over as I've been building it.

* The order of fields seems strange. Item 2 asks for the applicant's SSN even though they're applying for one. Items 11 through 13 asks if they've ever applied for an SSN before, the name on the card and the date of birth. All three of those should go together it seems and the SSN question is dependent on whether they applied before?
* Items 9 and 10 are inconsistent. It asks for the mother's name at birth but doesn't ask for the birth name of the father. I guess they're driving at maiden name but it's not clear. Seems there are a lot of ambiguous possibilities here like legal name changes of the father, surrogacy, unwed mothers who can't identify the father, etc.
* Item 4 asks for applicant's date of birth. In the instructions it notes that if the applicant is under 18 then the parent's SSNs are required. If that was absolutely true I'd be inclined to make these required fields based on the birth date value but then the instructions say "If you cannot obtain it checked Unknown" and there are the issues mentioned above as well so 🤷‍♂️ .
* There are some choose-one answers where it would be handy to have an easy way to display text or additional questions based on the answer, but it's cumbersome to do so using USFS.
  - For "Has the applicant ever applied for a Social Security Card" they can answer "Yes", "No", or "I don't know". If the answer is "Yes" more questions appear. For the other answers there are no additional questions but I suspect that the SSA may want to distinguish their answer anyway?
  - When the Citizenship answer is "Legal alien NOT allowed to work" or "Other" they need to provide further documentation, and it would be useful to be able to show more instructions under the answer and potentially some sort of checkbox saying "I have attached the documentation".

dmethvin-gov:
Here is the form.js with some of the issues above unresolved:
https://gist.github.com/dmethvin-gov/72127ae3c9a691169578688418164903

Issue #250, opened 2018-09-05
Create exemplar form: CMS40B-E


dmethvin-gov:
See discussion in #204.

Application for Enrollment in Medicare Part B (https://www.cms.gov/Medicare/CMS-Forms/CMS-Forms/Downloads/CMS40B-E.pdf)

dmethvin-gov:
In this form, there are several pages of "instructions" that need to be read carefully to determine if you should or could fill out the relatively simple form. Many if not most of them could be reduced to a set of questions that our form could use to prequalify someone, rather than make them read several pages of instructions. For example,

* You must already have Medicare Part A (otherwise you won't have a Medicare number!)
* If you do, enrollment for Medicare Part B is allowed:
  - for anyone during the *General Enrollment Period (GEP)* Jan 1 through March 31 of each year,
  - during your *Initial Enrollment Period (IEP)* which is from 3 months prior to 3 months after either your 65th birthday or your 25th month of disability (a 7-month window),
  - during an 8-month *Special Enrollment Period (SEP)* defined by several other conditions too incredibly esoteric to explain here, but determinable by Q&A

Would our process here be to try and incorporate some of that logic into the form? If so, are there PRA implications? Technically it doesn't have to be a part of the form, but given that this form will probably end up as an output in paper it might be useful to have an unofficial cover sheet that explains why the person qualifies so that if they go to a Medicare office it will be clear and they won't have to explain.

If we don't address such things this form is pretty easy to implement, but also not very interesting as an exemplar form since it can't be submitted or signed.

dmethvin-gov:
On the PRA aspect, it seems like answering questions in order to see if you qualify and should use the form might fall under the "data calculation" exclusion [mentioned here](https://obamawhitehouse.archives.gov/sites/default/files/omb/inforeg/memos/2014/appendix-data-search-tools-calculators.pdf) (Exclusions to the regulatory definition of “information” under Paperwork Reduction Act)

dmethvin-gov:
I seriously dislike the placement of buttons in GitHub.

dmethvin-gov:
Work is here until we figure out where to put and how to present:
https://gist.github.com/dmethvin-gov/39f5378080d45cb9bb35b6bd4aa57450

Issue #248, opened 2018-09-05
Create exemplar form: 4506-T


dmethvin-gov:
See discussion in #204.

@annekainicUSDS created this during the hackathon: https://github.com/annekainicUSDS/usfs-4506t

annekainicUSDS:
Included in this PR: https://github.com/usds/us-forms-system/pull/271

PR #271, opened 2018-09-18
Exemplar forms and documentation

### LINKED ISSUE #204: Identify exemplar forms
We need to show what is possible with the U.S. Forms System by providing a selection of exemplar forms.

To meet the goals we laid out in the original Project Charter, we should have at least three exemplars that cover the Easy, Medium, Hard spectrum.

dmethvin-gov:
Closes #204, #250, #251, #252, #246, #248

Still todo:
- [x] Add Anne's two forms (#246 and #248; add to the close list if they're put in this PR)
- [ ] Review forms and decide if we need additional comments or consistent style/treatment
- [ ] Create codesandbox.io versions of the final forms and update the links in the docs

## Types of changes

## Checklist:

- [x] ~~I added tests to verify a bug fix or new functionality.~~ (Examples tested with the starter app)

### DIFF
--- a/docs/examples/CMS40B-E--form.js
+++ b/docs/examples/CMS40B-E--form.js
@@ -1,0 +1,129 @@
+import Introduction from '../components/Introduction.jsx';
+
+const formConfig = {
+  title: 'Application for enrollment in Medicare Part B (medical insurance)',
+  subTitle: 'CMS40B-E',
+  formId: '',
+  urlPrefix: '/',
+  submitUrl: '',
+  introduction: Introduction,
+  confirmation: '',
+  defaultDefinitions: {},
+  preSubmitInfo: {
+    notice: 'I hope you did not lie!',
+    required: true,
+    label: 'I swear this information is correct to the best of my knowledge.',
+    field: 'signed',
+    error: 'You must check the box to submit this form and apply for Medicare Part B'
+  },
+  chapters: {
+    CP1: {
+      pages: {
+        PG1: {
+          path: 'verification',
+          title: 'Enrollment verification',
+          uiSchema: {
+            signUp: {
+              'ui:title': 'I wish to apply for Medicare Part B',
+              'ui:options': {
+                expandUnderClassNames: 'schemaform-expandUnder-indent'
+              }
+            },
+            'view:preQualification': {
+              'ui:description': 'To apply for Medicare Part B, you must have already applied for and recieved Medicare Part A. If you do not have Part A and want to sign up, please contact Social Security at 1-800-772-1213. TTY users should call 1-800-325-0778.',
+              'ui:options': {
+                expandUnder: 'signUp'
+              }
+            },
+            medicareNumber: {
+              'ui:title': 'Medicare Number',
+              'ui:description': 'Your Part A paperwork has a Medicare number in the form XXXX-XXX-XXXX consisting only of letters and digits, please enter it below.',
+              'ui:errorMessages': {
+                pattern: 'Your Medicare number should be in the form XXXX-XXX-XXXX, letters and digits separated by dashes'
+              }
+            }
+          },
+          schema: {
+            type: 'object',
+            required: [ 'signUp', 'medicareNumber' ],
+            properties: {
+              signUp: {
+                type: 'boolean'
+              },
+              'view:preQualification': {
+                type: 'object',
+                properties: {}
+              },
+              medicareNumber: {
+                type: 'string',
+                pattern: '[0-9a-zA-z]{4}-[0-9a-zA-z]{3}-[0-9a-zA-z]{4}'
+              }
+            }
+          }
+        },
+        PG2: {
+          path: 'enrollee-info',
+          title: 'Enrollee information',
+          uiSchema: {
+            zip: {
+              'ui:errorMessages': {
+                pattern: 'Please enter a 5-digit zip code or 9-digit zip+4 code'
+              }
+            },
+            phone: {
+              'ui:errorMessages': {
+                minLength: 'Please enter a phone number, including area code'
+              }
+            }
+          },
+          schema: {
+            type: 'object',
+            required: [ 'name', 'address', 'city', 'state', 'zip', 'phone' ],
+            properties: {
+              name: {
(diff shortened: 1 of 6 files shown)


docs/examples/CMS40B-E--form.js
```diff
@@ -0,0 +1,129 @@
...
+              }
+            }
+          },
+          schema: {
+            type: 'object',
+            required: [ 'name', 'address', 'city', 'state', 'zip', 'phone' ],
+            properties: {
+              name: {
+                type: 'string',
+                title: 'Full name (Last, First, Middle)',
```
annekainicUSDS: This is very interesting, usually I use `ui:title` under `uiSchema`, which you used above. I don't think it matters, seems like they're both treated the same, but I wonder if we should consistently use one over the other?

docs/examples/DS-82--form.js
```diff
@@ -0,0 +1,262 @@
+
```
annekainicUSDS: Extra line nit!

docs/examples/DS-82--form.js
```diff
@@ -0,0 +1,262 @@
...
+              'ui:errorMessages': {
+                minLength: 'Please enter a phone number, including area code'
+              }
+            },
+            ssn: {
+              'ui:errorMessages': {
+                pattern: 'Please enter a social security number in the form 000-00-0000'
+              }
+            },
+            dob: currentOrPastDate('Date of birth')
```
annekainicUSDS: I assume this is the documentation for version 1.1.0, because the definition files have been changed since then, but not yet released.

docs/examples/README.md
```diff
@@ -0,0 +1,30 @@
+# Form.js examples
+
+The `formConfig` object configures most of the form inputs, pages, chapters, and settings in a USFS project. In the [starter app](https://github.com/usds/us-forms-system-starter-app/), this file is named `form.js` and is stored [in the config directory](https://github.com/usds/us-forms-system-starter-app/tree/3617106dfea7c42f5c397f96f24fb906c8e2879d/js/config).
+
+To use one of the example files with the starter app, rename it to `form.js` and replace the file in a locally installed version of the starter app. Instructions for setting up the starter app are [in the documentation](https://github.com/usds/us-forms-system/blob/master/docs/getting-started/creating-a-new-application-with-the-us-forms-system-starter-app.md). Online versions of the forms are also available on [codesandbox.io](https://codesandbox.io) that can be used for experimentation. The examples available are listed below.
+
+### CMS40B-E: Appication for Enrollment in Medicare Part B
+* [Original PDF form](https://www.cms.gov/Medicare/CMS-Forms/CMS-Forms/Downloads/CMS40B-E.pdf)
+* [form.js file](./CMS40B-E--form.js)
+* [codesandbox.io form using the starter app]()
```
annekainicUSDS: Did you mean to still add links for all the sandbox references?

docs/examples/SS-5--form.js
```diff
@@ -0,0 +1,280 @@
+import { currentOrPastDateConfig } from 'us-forms-system/lib/js/definitions/currentOrPastDate';
```
annekainicUSDS: This use of `currentOrPastDateConfig` uses the new version of this definition, so perhaps we want to be consistent with how it was used in the other file as well.

docs/examples/CMS40B-E--form.js
dmethvin-gov: It works with either. In places where it saved having an entry at all in `uiSchema` for that field I put it into `schema` to shorten things up. We can use `ui:title` consistently, although it does point out the verbosity of the current setup.

dmethvin-gov: Looking at some of these forms I see I could have saved even more than I did by going all-`title` in more places, like [here](https://github.com/usds/us-forms-system/pull/271/files#diff-311ffb3b6f74e31e3c3b94c1d2387eedR19). It really shows how much noisier it is to have two sections instead of one.

annekainicUSDS: Yes, good point. In that case totally fine to leave `title` for now and then we can revisit this when we reimagine the config :)

docs/examples/DS-82--form.js
annekainicUSDS: Yes, I agree it would be nice not to have to go back. Perhaps we should decide definitely when we're releasing the new definitions; if it's going to be next week we might as well update the docs to reflect that now.

annekainicUSDS:
@dmethvin-gov Do you want me to add the forms I had built to this branch? Or should I do that after this has been merged? I assume you'll need to add them to the codesandbox for me?

annekainicUSDS:
@dmethvin-gov I created a branch off of yours, and then tried to submit a PR into your branch, but it's not allowing me, because I think you're working off a fork. So here's my PR: https://github.com/usds/us-forms-system/pull/281. I don't know what's the best way to incorporate my changes, either merging them after your merge your PR, or you can just copy my changes into your branch.

docs/examples/DS-82--form.js
dmethvin-gov: Since we didn't pull in the definition changes for 1.2.0 I am going back to the old style.

dmethvin-gov:
I've spent a bunch of time today trying to get the codesandbox examples to work. There are some issues with the dependency setups we have in the package.json files for both USFS and the starter app that make them not work correctly with the sandbox. Instead of blocking this PR I think it is better to land this and straighten out the dependency issues separately. Does that sound like a plan?

The short story for the main dependencies issue is that if you are running a production build you still need to pull in a bunch of the devDependencies in order to actually build the production code. After that, you actually have the built files and can run without things like webpack. That plus a few other minor issues like USWDS being in the wrong place confuse codesandbox right now. I'll put together a PR once I understand how to fix it all.

annekainicUSDS:
Yes I think it’s fine to fix those issues in a separate PR, sounds like a
pain, sorry about that!

dmethvin-gov:
@annekainicUSDS if you're good with it for now can you approve? I took care of the extra line and am using the 1.1.0 definitions now.

PR #289, opened 2018-09-27
Remove vets.gov-isms

### LINKED ISSUE #211: Eliminate vets.gov-isms in form submit and file upload

dmethvin-gov:
Fixes #211

I tried to keep the commits subject-specific so they're easier to sort out.

- [x] TODO: Update documentation
- [ ] TODO: Commit built files (I didn't commit them yet to keep the diff clean)

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] ~~I added tests to verify a bug fix or new functionality.~~ Removed now-irrelevant tests

### DIFF
--- a/src/js/actions.js
+++ b/src/js/actions.js
@@ -141,9 +141,7 @@ export function submitForm(formConfig, form) {
         ? formConfig.transformForSubmit(formConfig, form)
         : transformForSubmit(formConfig, form);
 
-      promise = formConfig.submitToUrl
-        ? formConfig.submitToUrl(body, formConfig.submitUrl, recordEvent)
-        : submitToUrl(body, formConfig.submitUrl, recordEvent);
+      promise = submitToUrl(body, formConfig.submitUrl, recordEvent);
     }
 
     return promise
--- a/src/js/actions.js
+++ b/src/js/actions.js
@@ -76,7 +76,7 @@ export function setViewedPages(pageKeys) {
   };
 }
 
-function submitToUrl(body, submitUrl, recordEvent) {
+export function submitToUrl(body, submitUrl, recordEvent) {
   return new Promise((resolve, reject) => {
     const req = new XMLHttpRequest();
     req.open('POST', submitUrl);
@@ -91,10 +91,10 @@ function submitToUrl(body, submitUrl, recordEvent) {
       } else {
         let error;
         if (req.status === 429) {
-          error = new Error(`vets_throttled_error: ${req.statusText}`);
+          error = new Error(`throttled_error: ${req.statusText}`);
           error.extra = parseInt(req.getResponseHeader('x-ratelimit-reset'), 10);
         } else {
-          error = new Error(`vets_server_error: ${req.statusText}`);
+          error = new Error(`server_error: ${req.statusText}`);
         }
         error.statusText = req.statusText;
         reject(error);
@@ -102,30 +102,24 @@ function submitToUrl(body, submitUrl, recordEvent) {
     });
 
     req.addEventListener('error', () => {
-      const error = new Error('vets_client_error: Network request failed');
+      const error = new Error('client_error: Network request failed');
       error.statusText = req.statusText;
       reject(error);
     });
 
     req.addEventListener('abort', () => {
-      const error = new Error('vets_client_error: Request aborted');
+      const error = new Error('client_error: Request aborted');
       error.statusText = req.statusText;
       reject(error);
     });
 
     req.addEventListener('timeout', () => {
-      const error = new Error('vets_client_error: Request timed out');
+      const error = new Error('client_error: Request timed out');
       error.statusText = req.statusText;
       reject(error);
     });
 
-    req.setRequestHeader('X-Key-Inflection', 'camel');
     req.setRequestHeader('Content-Type', 'application/json');
-
-    const userToken = _.get('sessionStorage.userToken', window);
-    if (userToken) {
-      req.setRequestHeader('Authorization', `Token token=${userToken}`);
-    }
 
     req.send(body);
   });
@@ -148,7 +142,9 @@ export function submitForm(formConfig, form) {
         ? formConfig.transformForSubmit(formConfig, form)
         : transformForSubmit(formConfig, form);
 
-      promise = submitToUrl(body, formConfig.submitUrl, recordEvent);
+      promise = formConfig.submitToUrl
+        ? formConfig.submitToUrl(body, formConfig.submitUrl, recordEvent)
+        : submitToUrl(body, formConfig.submitUrl, recordEvent);
     }
 
     return promise
@@ -158,9 +154,9 @@ export function submitForm(formConfig, form) {
         const error = errorReceived instanceof Error ? errorReceived : new Error(errorReceived);
         const errorMessage = String(error.message);
         let errorType = 'clientError';
-        if (errorMessage.startsWith('vets_throttled_error')) {
+        if (errorMessage.startsWith('throttled_error')) {
           errorType = 'throttledError';
-        } else if (errorMessage.startsWith('vets_server_error')) {
+        } else if (errorMessage.startsWith('server_error')) {
           errorType = 'serverError';
         }
         recordEvent({ event: 'form-submit-error', error, errorType });
@@ -229,8 +225,6 @@ export function uploadFile(file, uiOptions, onProgress, onChange, onError) {
           name: file.name,
           errorMessage
         });
-        // Commenting until this is removed by a PR for #211
-        //      Raven.captureMessage(`vets_upload_error: ${req.statusText}`);
         onError();
       }
     });
@@ -241,12 +235,6 @@ export function uploadFile(file, uiOptions, onProgress, onChange, onError) {
         name: file.name,
         errorMessage
       });
-      // Commenting until this is removed by a PR for #211
-      //      Raven.captureMessage(`vets_upload_error: ${errorMessage}`, {
-      //        extra: {
-      //          statusText: req.statusText
-      //        }
-      //      });
       onError();
     });
 
--- a/src/js/actions.js
+++ b/src/js/actions.js
@@ -1,4 +1,3 @@
-import _ from './utilities/data';
 import { transformForSubmit } from './helpers';
 import { timeFromNow } from './utilities/date';
 
--- a/src/js/definitions/phone.js
+++ b/src/js/definitions/phone.js
@@ -22,7 +22,7 @@ function uiSchema(title = 'Phone') {
       pattern: 'Phone numbers must be 10 digits'
     },
     'ui:options': {
-      widgetClassNames: 'home-phone va-input-medium-large',
+      widgetClassNames: 'home-phone usfs-input-medium-large',
       inputType: 'tel'
     }
   };
--- a/src/js/helpers.js
+++ b/src/js/helpers.js
@@ -242,22 +242,12 @@ export function filterInactivePageData(inactivePages, activePages, form) {
 }
 
 export function stringifyFormReplacer(key, value) {
-  // an object with country is an address
-  if (value && typeof value.country !== 'undefined' &&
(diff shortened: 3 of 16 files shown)


src/js/actions.js
```diff
@@ -76,7 +75,7 @@ export function setViewedPages(pageKeys) {
   };
 }
 
-function submitToUrl(body, submitUrl, recordEvent) {
+export function submitToUrl(body, submitUrl, recordEvent) {
```
dmethvin-gov: This is pretty generic now but potentially could be used by someone with plain-vanilla needs.

src/js/actions.js
```diff
@@ -148,7 +141,9 @@ export function submitForm(formConfig, form) {
         ? formConfig.transformForSubmit(formConfig, form)
         : transformForSubmit(formConfig, form);
 
-      promise = submitToUrl(body, formConfig.submitUrl, recordEvent);
+      promise = formConfig.submitToUrl
```
dmethvin-gov: I've added a `formConfig.submitToUrl` option that needs to be documented. The default just removes empty objects and arrays (plus uploads which may possibly go).

src/js/definitions/phone.js
```diff
@@ -22,7 +22,7 @@ function uiSchema(title = 'Phone') {
       pattern: 'Phone numbers must be 10 digits'
     },
     'ui:options': {
-      widgetClassNames: 'home-phone va-input-medium-large',
```
dmethvin-gov: The classes across the SCSS and JS files were just renamed. I didn't see this particular class defined anywhere so it may be missing?

test/js/actions.unit.spec.js
```diff
@@ -129,31 +129,6 @@ describe('Schemaform actions:', () => {
 
       return promise;
     });
-    it('should submit with auth header', () => {
```
dmethvin-gov: This should have been removed earlier since we don't support users.

test/js/helpers.unit.spec.js
```diff
@@ -425,28 +425,6 @@ describe('Schemaform helpers:', () => {
         anotherField: 'testing3'
       });
     });
-    it('should remove empty addresses', () => {
```
dmethvin-gov: This functionality is now removed.

test/js/helpers.unit.spec.js
```diff
@@ -504,30 +482,6 @@ describe('Schemaform helpers:', () => {
       expect(output.arrayField.length).to.equal(1);
       expect(output.emptyArray).to.be.undefined;
     });
-    it('should convert autosuggest field to id', () => {
```
dmethvin-gov: This functionality has been removed.

test/js/review/SubmitButtons.unit.spec.jsx
```diff
@@ -41,18 +41,6 @@ describe('Schemaform review: <SubmitButtons>', () => {
     expect(tree.everySubTree('ProgressButton')[1].props.buttonText).to.equal('Submitted');
     expect(tree.everySubTree('ProgressButton')[1].props.disabled).to.be.true;
   });
-  it('should render error in dev mode', () => {
```
dmethvin-gov: This functionality has been removed.

src/js/helpers.js
```diff
@@ -242,11 +242,6 @@ export function filterInactivePageData(inactivePages, activePages, form) {
 }
 
 export function stringifyFormReplacer(key, value) {
-  // an object with country is an address
```
dmethvin-gov: Vets.gov will need to add back the special cases here if they need them.

src/js/actions.js
dmethvin-gov: So now there is  the existing `formConfig.submit` which lets the form creator completely replace the submit process incuding any pre-form massaging, plus the new `formConfig.submitToUrl` for just the network part.

annekainicUSDS: So if I'm understanding this correctly, there are now 3 options in the config for submit actions:
1. `submit`: a function to handle the form submission
2. `submitToUrl`: a function to handle the specific part of the submission where data is actually posted, and to log responses back
3. `submitUrl`: a string of the url the data is being submitted to

Is it actually necessary to have `submitToUrl` (2) as an option? It seems like `submitToUrl()` is just a subset of the general `submitForm()` function, and some of the logic in `submitForm()` is actually dependent on logic in `submitToUrl()`, with the checks for the errors. 

Are these functions actually wholly distinct from one another? Can we imagine a world where someone wants to use the default option for one, but write a custom option for the other?  If someone wanted to rewrite both, I think all we would need is the `submit` option in the config, and someone would have to pass a function to that that covers both `submitForm()` and `submitToUrl()` functionality. What do you think?

src/js/helpers.js
```diff
@@ -255,11 +250,6 @@ export function stringifyFormReplacer(key, value) {
       return undefined;
     }
 
-    // autosuggest widgets save value and label info, but we should just return the value
```
annekainicUSDS: I don't know if this is a special case as much as it is the way our widget is written; we may not be able to remove this, because otherwise the wrong data will be submitted for every instance of the autosuggest field. A more permanent fix might be changing what the widget returns, but for now I think this needs to stay?

src/js/definitions/phone.js
annekainicUSDS: Yes I don't see this class defined anywhere in our codebase, so I think we can just remove it from here entirely, if it's not actually adding any styling.

src/js/review/SubmitButtons.jsx
```diff
@@ -45,8 +45,8 @@ export default function SubmitButtons(props) {
     submitMessage = (
       <div className="usa-alert usa-alert-error schemaform-failure-alert">
         <div className="usa-alert-body">
-          <p className="schemaform-warning-header"><strong>We’re sorry, there was an error connecting to Vets.gov.</strong></p>
-          <p>Please check your Internet connection and try again. If the problem persists, please contact the Vets.gov Help Desk.</p>
+          <p className="schemaform-warning-header"><strong>We’re sorry, there was an error connecting to the server.</strong></p>
```
annekainicUSDS: Does this change provide a way for Vets.gov to still use the original language? I don't think there's an option in the config for overwriting the default language here, as there is for the presubmit info.

test/js/helpers.unit.spec.js
annekainicUSDS: I thought we still had autosuggest functionality?

src/js/actions.js
dmethvin-gov: I'd be fine with just having `formConfig.submit` be the only option and exposing these to either be used as-is or as templates for building custom code.

src/js/helpers.js
dmethvin-gov: Does the "we should" in the comment represent what every library user would want from the widget? (In other words, nobody wants the labels.) If so we should change the widget so this transformation isn't necessary. If this is more about how vets.gov would like to see the info at their back-end, they can add their own custom transformation. Either way it's a breaking change.

src/js/review/SubmitButtons.jsx
dmethvin-gov: There are a lot of things in this file including styling, text , and error messages that someone might want to configure. There is even some logic, and someone might want to put the retry logic back for example. I think this whole file should end up in the starter app because there's just too much to try and configure via properties in `formConfig`.

test/js/helpers.unit.spec.js
dmethvin-gov: We do, but this was specifically testing whether autosuggest *data* is transformed and doesn't test the widget. As far as I can tell there are not any tests for the autosuggest widget functioning.

src/js/helpers.js
```diff
@@ -255,11 +250,6 @@ export function stringifyFormReplacer(key, value) {
...
-    if (value.widget === 'autosuggest') {
-      return value.id;
-    }
-
     // Exclude file data
```
dmethvin-gov: I actually had questions about whether to remove this as well. I wasn't sure exactly how the uploading works. The data isn't attached to the form so how does the submitted form get associated with the uploaded file when the backend wants to do something with them both?

src/js/actions.js
annekainicUSDS: I think that might make things simpler!

src/js/helpers.js
annekainicUSDS: Good question; I'm not sure! I guess we don't know until we get other users. Given our only current user wants only the values, I would advocate for keeping this transformation in here, or at least changing the widget so this transformation isn't necessary. But curious to hear what you think!

src/js/review/SubmitButtons.jsx
annekainicUSDS: Good point, there should be an easy way to swap out the entire component if you want something different.

src/js/helpers.js
annekainicUSDS: I don't think there's uploading here? This is for an autosuggest field, not file upload. I assume it's just data being populated?

dmethvin-gov: I meant the block under the `// Exclude file data` comment that follows it. This is another duck-typed modification but I think it may be making vets.gov assumptions that don't apply generally.

src/js/helpers.js
dmethvin-gov: They can still have the values the way they want them, by doing their own modification of the data before submit. It sounds like vets.gov will need that function already if they still need that special case for addresses. 

Think about it this way, if we add more widgets would we add special case code for those widgets here as well? It doesn't seem so. Either the widget would produce the "right" format in the form or we'd ask users to transform it themselves before submitting if they prefer a different format.

src/js/helpers.js
annekainicUSDS: Looking at this now, I think I realize why this was added: for our address fields we default the country to `USA`. If the user is able to skip over entering the address (i.e., it's not required), the `country` field will be submitted with the form without any of the other address fields, which doesn't make a lot of sense. So in those cases, we just remove those fields.

I know we removed the address definition from our codebase that would include the logic to default the country, so this change doesn't break anything within USFS, but certainly the Vets.gov team will want to to add this special case back in. Is the only way for them to add this case back in to replace the entire `transformForSubmit` function? Or at least the `stringifyFormReplacer` part that is passed to `transformForSubmit`?

src/js/definitions/phone.js
annekainicUSDS: Minor nit, but I think this class, `usfs-input-medium-large`, can be removed.

dmethvin-gov:
> Is the only way for them to add this case back in to replace the entire `transformForSubmit` function? Or at least the `stringifyFormReplacer` part that is passed to `transformForSubmit`?

That certainly looks like the case today. It's never said anywhere but I guess a widget is supposed to be in a schema-consistent state at any point while the form is being filled since it's subject to the (supposedly final) JSON schema. The exceptions are cleaned up by `transformForSubmit` but those would have to be conditions that are not constrained by that particular form's JSON schema, otherwise it would have thrown an error during validation.

It seems like the right place for these sort of fixups are some type of hook in the widget definition rather than a global function that knows about the union of all widgets. The library can't know all the widgets it might be used with or whether a particular back-end needs the data to be transformed. For now though, any library user would need to write their own custom `transformForSubmit` if they had any widgets that needed fixups.

PR #291, opened 2018-09-28
Allow year definition to include future years

### LINKED ISSUE #221: Change `year.js` to accept future years
Right now the `year` definition only accepts current or past years. There may be use cases where a future year is needed. Allow future years to be accepted, and create a separate definition for `currentOrPastYear`.

dmethvin-gov:
Fixes #221

There were a few ways to fix this. Since this is just an integer I decided to limit the input to the year 2100 rather than creating a custom validation function.

The upside of doing that is that if at some point someone wants to enter a year of 2116 or something they will get a specific error message saying the maximum is 2100 and know what the problem is. (The typical error for the custom validator is "Please enter a valid year" which tells you less.)

## Types of changes

## Checklist:


dmethvin:
I think in this case it is a bug fix, since it wasn't working as most people would expect. We can add a note to the release notes about this change, or we could only land it in master and not the 1.x branch if vets.gov doesn't want it.

PR #294, opened 2018-10-02
Use link checker for markdown files, add to lint tasks

### LINKED ISSUE #145: Add link checker to CI
Per #143 I think we should be able to add https://github.com/tcort/markdown-link-check pretty easily. Rather than make it part of `npm run build` I'm inclined to just make it part of the CI for PRs and merges to master. When #97 is resolved we'll have a commit hook that runs build to ensure that all the built files are in sync with the source, and I'm concerned that checking all the links on each build would take too long.

dmethvin-gov:
Fixes #145

The link checking takes a long time when the file is big or there are many external links, so I'm only checking the markdown files actually changed in the commit.

I also added a `npm run lint:md-full` command that lints all the files in case there is bit rot. That found two broken links and I fixed them here.

## Types of changes

## Checklist:


dmethvin-gov:
One complication I thought of at lunch is that the PR could consist of multiple commits and this only checks the last one. When the commit is squashed to merge any problem should be detected then but ideally it would be nice to check all the commits since the shared base commit. Not sure how easy that is to find (it won't always be `master`) but I'll look.

dmethvin-gov:
Just noting here that if you know the branch it's possible to find the common commit using `git merge-base HEAD master` but I don't think we can assume the branch.

annekainicUSDS:
Hm, this issue of only checking the last commit and potentially missing issues in earlier commits would also affect CI, right? I don't love the idea that we couldn't 100% depend on CI to catch issues before merging a PR. I'd like to find a better solution to this if possible. I'll dig around.

dmethvin:
If we squash merges then there will only be one commit. The only time we run into the multiple commits problem is on PRs. I just didn't have a good way to only check the files in the PR and a full check is too expensive. If there were some broken links in an earlier commit on a PR then it should be caught on the squashed merged. There is also the command line option to fully check locally which is probably a good idea when a LOT of docs files are changing but again too expensive to do as part of CI.

Issue #66, opened 2018-06-06
Discovery on best way to import USWDS images and fonts


annekainicUSDS:
Right now, Vets.gov and the VA design system imports the images and fonts from USWDS directly, by coping all of the files into an assets directory. This is what I've done temporarily for USWDS as well. There may be a better solution though; look into webpack handling this potentially.

jcmeloni-usds:
So (ignore if this is too much of a tangent) is uswds in the package.json not really fully useful if we have to manually copy things anyway?  Or is it that the package puts things in weird places and it's too much work to remap where those places are so we just manually copy them over?

annekainicUSDS:
These are good questions that I need to answer when looking into this ticket more :) For expediency I just copied what vets.gov was doing to get the starter-app working, but definitely want to find a better long-term solution

jcmeloni-usds:
haha oops, my bad?

annekainicUSDS:
That's okay! Wasn't clear from the description

annekainicUSDS:
Another problem to look into here: One of the fonts that is used in our forms is imported in a font faces file in the VA design system: https://github.com/department-of-veterans-affairs/design-system/blob/master/src/sass/base/_b-font-faces.scss. This font is copies into vets-website under the `assets/` directory, and then this directory is redirected to `/` in the build script: https://github.com/department-of-veterans-affairs/vets-website/blob/master/script/build.js#L526. When the stylesheet looks for the font, it will look at `/font/...`. This is a problem because in us-forms-system we're not redirecting our fonts to the base path, so the url that is specified in the font-faces file from VA design can't be found.

As a temporary measure, I've created a different version of the font faces file that changes the path to be relative to the directory it's in, and then copied those fonts into `lib/fonts`, but would like to find a better long-term solution for this. PR that makes these changes is here: https://github.com/usds/us-forms-system/pull/81

annekainicUSDS:
I'd like to keep this ticket open since it's slightly different from the work I've been doing to refactor the scss (haven't specifically addressed USWDS issues yet, have been focusing on pulling out VA design system styles).

jcmeloni-usds:
@annekainicUSDS @fatima-gov Is there a tl;dr on what's happened with this or is the reference to #152 the most accurate? Thanks!

annekainicUSDS:
[...]

PR #295, opened 2018-10-04
[WIP] Remove build artifacts from the repo

### LINKED ISSUE #218: Strategy for committing build artifacts

dmethvin-gov:
Fixes #218 

This is a breaking change since potentially some people might be expecting the `/lib` contents to be there on each build.  A couple of things:

* I have questions about where the images and fonts should go, they were in `/lib` but aren't copied there? I think our build process needs to copy anything we require or we need to figure out a better way to reference them. (This is the subject of #66 I think but I wanted to be sure I understood all the issues here.)

* As far as leaving things out, should we remove our eslint and babel configs from the published package? It depends on whether we expect people to transpile our code, there are a bunch of pro/cons but I was thinking it might be good to leave it in for now.

Once those are resolved I think it should be possible to land this. I ran a `npm publish --dry-run` and it seemed to put in the right files.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] I've reviewed the [guidelines for contributing to this repository](../../CONTRIBUTING.md#how-to-contribute-to-us-forms-system).
- [ ] I've checked style and lint with `npm run lint`.
- [ ] All tests pass locally with `npm test`.


annekainicUSDS:
@dmethvin-gov to answer your questions:

1. Yes, those images and fonts were manually copied directly into the `/lib` directory once. Not at all ideal, but that's what that ticket you referenced is meant to look into. I assume we would need to address that before this PR could be merged?

2. I think we decided at the beginning of the project to not assume that users will do their own transpilation, and for us to include the transpiled files. If we aren't committing these files, when are they going to be created and how will they be used? I'm a little unclear on how this change to remove build artifacts completely works with our current assumptions about how users import files from USFS into their own repos, and what would need to change.

dmethvin-gov:
On 1, if that's the way we want it to stay we can just copy them into `/lib` when `npm run build` happens. I was trying to figure out a way to avoid pulling in the entire USWDS since it's a 181KB CSS file but I'm not sure how to do that since our caller may need things we don't use directly. We've got USWDS listed as a `peerDependency` but if we are packaging it all in our lib it's really not a dependency of our caller, but just us.

On 2, all of the files are always going to be built and linted on each commit to master or on PRs, since they're needed to run tests. It's just a question of whether `/lib` is saved or not after that is done. Right now we're saving all that to the repo, but with this PR we'd just save it when we publish. The new `.npmignore` takes care of that.

dmethvin-gov:
Looking at this some more I think we should have this PR resolve #66 along with #218. Some thoughts about it:

* If we copy just some of the USWDS `/dist` files we need into locations outside `node_modules` and check them into our repo and/or publish our package with them, we should also be copying the license info for it and mention their license in ours. 
* We can avoid those complications if we could use the files directly out of `node_modules`, at that point it could be a true `peerDependency`. Right now the starter app is using what it does use in USWDS through us, which means even patch versions of USWDS would require us to cut a new release.
* I wonder how others are using USWDS via npm, and if so what does their setup look like? The [installation instructions](https://designsystem.digital.gov/documentation/developers/#installation) seem to encourage npm and don't talk about it being copied.

package.json
```diff
@@ -8,16 +8,19 @@
     "test": "test"
   },
   "scripts": {
-    "build": "rm -rf lib/{js,css} && npx babel src/js --source-maps --out-dir lib/js && npm run compile-sass && npm run autoprefix",
-    "compile-sass": "node-sass --source-map lib/css/ --source-map-embed false --include-path src/scss/styles.scss src/scss/styles.scss -o lib/css/",
     "autoprefix": "npx postcss lib/css/styles.css --use autoprefixer -map -d lib/css/",
+    "build": "npm run compile-js && npm run compile-sass && npm run autoprefix && npm run copy-uswds -s",
```
annekainicUSDS: We've removed the part where we clear out the `lib/` directory first before adding the compiling files. Is that not something we still need to do? To make sure that files we may have deleted are being deleted?

package.json
```diff
@@ -8,16 +8,19 @@
...
+    "ci": "npm run clean -s && npm run lint && npm run build",
```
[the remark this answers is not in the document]
dmethvin-gov: Ugh, we do but I messed up fixing the merge conflict. This was exactly one of those situations where the built files updated causing a merge conflict, . I can add it back. Thanks for catching it!

package.json
dmethvin-gov: For CI it's nice to be completely sure it's clean but I wasn't sure whether it was really needed when you're building over and over locally. Files removed from the project are relatively rare. I can put it in here as well though if you think we should have it.

annekainicUSDS: I think it can't hurt! Might as well include it to be extra sure it's reflective of reality.

package.json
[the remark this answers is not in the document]
annekainicUSDS: Ah gotcha!

dmethvin-gov:
OK, I moved all the steps into `build`, including `lint`. Later we may want to add some less-costly shortcuts because it takes a while to run.

dmethvin-gov:
Once we remove the build artifacts we could also revisit the use of a push hook or commit hook which we couldn't use before.

PR #300, opened 2018-10-19
Revert "Use link checker for markdown files, add to lint tasks (#294)"

dmethvin-gov:
This works locally but not in CircleCI. I'm going to back it out until I figure out why.


dmethvin-gov:
This appears it might be some sort of bug where it works in node 10 (which I'm using) but not node 8 (which CircleCI uses for testing). I can obviously upgrade to 10 but it really should work in 8 as well. The error being thrown is "path must be a string or buffer" which is internal to node. https://circleci.com/gh/dmethvin-gov/us-forms-system/159

I'll keep looking.

dmethvin-gov:
Closing in favor of #301

PR #301, opened 2018-10-19
Work around differences in xargs on Linux vs Mac

dmethvin-gov:
xargs works differently on Mac vs Linux which causes it to call the link checker differently. This uses a different approach to avoid the difference.


dmethvin-gov:
Haha, you can always count on me to find a completely bizarre bug once in a while!

This markdown scanner is a bit limited and only allows one file to be passed to it at a time. Soooo, I have to call it once for each file that changed. Since I have to generate the list of files via some `git` commands, it was easiest to run through `xargs` which can call a command for each line in its input.

MacOS is based on Unix but the CircleCI setup is Linux-based. Although they both have `xargs` they appear to work a bit differently. I was using the `-L` option that pulls one per line but there seems to be a special case where there is a blank line in the input. The MacOS `xargs` ignores it so there's no problem, but the Linux `xargs` passes something strange as input to the command which caused the markdown parser to die within Node with that strange message.

PR #312, opened 2018-10-30
[For discussion] Create build artifacts in separate repo

dmethvin-gov:
Since we're not storing the build artifacts anymore, it is not quite as convenient to reference a specific branch, or commit for testing (not production!) with an external app. This PR is a proof-of-concept for one way to store these builds using a second Github repo. 

The basic approach is to have CircleCI take the build artifacts generated in `lib/` and commit them to a new repo. Then it's possible to reference built files in that repo using either a commit hash or a branch name. I don't have admin access to create the `usds/usfs-builds` repo so I created one under my own account, along with SSH keys that are locked down to just that new repo.  [Here's what a commit looks like.](https://github.com/dmethvin-gov/usfs-builds/commits/builds--218)

One other thing we might consider doing in this step is actually running the tests with the built files. None of our tests are currently running with the built files, they are being compiled via mocha during testing and then discarded. Although you could argue that it *should* compile to the same result it would be reassuring if we actually tested with the files we plan to distribute.

## Types of changes

## Checklist:

- [ ] I've reviewed the [guidelines for contributing to this repository](../CONTRIBUTING.md#how-to-contribute-to-us-forms-system).
- [ ] I've checked style and lint with `npm run lint`.
- [ ] All tests pass locally with `npm test`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,8 +28,28 @@ jobs:
     steps:
       - attach_workspace:
           at: ~/repo
-      - run: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > ~/repo/.npmrc
-      - run: npm publish
+      - run: |
+          echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > ~/repo/.npmrc;
+          npm publish
+
+  artifacts:
+    <<: *common
+    steps:
+      - attach_workspace:
+          at: ~/repo
+      - add_ssh_keys:
+          fingerprints:
+            - "57:4b:f6:31:3f:dc:80:fd:ee:b4:e4:ba:0f:98:20:22"
+      - run: |
+          cd ~/repo
+          git config user.email "USFS-bot@usds.gov"
+          git config user.name "USFS Bot"
+          echo -e "\nHost github.com\n\tStrictHostKeyChecking no\n" >> ~/.ssh/config
+          grep -v lib/ .gitignore >.gitignore-new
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -28,8 +28,28 @@ jobs:
...
+            - "57:4b:f6:31:3f:dc:80:fd:ee:b4:e4:ba:0f:98:20:22"
+      - run: |
+          cd ~/repo
+          git config user.email "USFS-bot@usds.gov"
+          git config user.name "USFS Bot"
+          echo -e "\nHost github.com\n\tStrictHostKeyChecking no\n" >> ~/.ssh/config
+          grep -v lib/ .gitignore >.gitignore-new
+          mv .gitignore-new .gitignore
+          git add lib/
+          git commit -m "CircleCI $CIRCLE_BRANCH -- add built files in lib/"
```
annekainicUSDS: What does $CIRCLE_BRANCH refer to? Is that created automatically?

.circleci/config.yml
```diff
@@ -42,3 +62,6 @@ workflows:
           filters:
             branches:
               only: production
+      - artifacts:
```
annekainicUSDS: Do we want this job to fall under the `build_and_deploy` category? Wouldn't this happen automatically at every commit that is pushed up to a remote branch in this repo?

dmethvin-gov:
I don't think we need this at the moment. If we split the library components in order to use them separately I think it might be easier to use a lerna monorepo for this.

PR #328, opened 2019-01-04
Update dependencies to remove warnings when using React 16

jbalboni:
On VA.gov we're updating to React 16 and USFS has some peer dependency rules that cause some warnings for us when upgrading. There's also one issue with a direct dependency, recompose, so that library is updated to the latest version.

So far in our VA.gov testing, our forms work without specific changes for React 16.

The recompose update is not a major version update, but I did come across a break change in the tests, in https://github.com/acdlite/recompose/releases/tag/v0.25.0. I made some test updates for that change (though I'm not clear on exactly why their change needed this particular change in the tests).

## Types of changes

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,8 +84,8 @@
     "uswds": "^1.6.3"
   },
   "dependencies": {
-    "classnames": "^2.2.5",
     "@department-of-veterans-affairs/react-jsonschema-form": "^1.0.0",
+    "classnames": "^2.2.5",
     "downshift": "^1.22.5",
     "fast-levenshtein": "^2.0.6",
     "history": "3",
@@ -95,7 +95,7 @@
     "react-router": "3",
     "react-scroll": "^1.7.9",
     "react-transition-group": "^1.1.2",
-    "recompose": "^0.21.2",
+    "recompose": "^0.30.0",
     "redux": "^3.5.2",
     "reselect": "^2.5.4"
   }
--- a/package.json
+++ b/package.json
@@ -79,9 +79,9 @@
     "vets-json-schema": "git+https://github.com/department-of-veterans-affairs/vets-json-schema.git"
   },
   "peerDependencies": {
-    "react": "^15.5.4",
-    "react-dom": "^15.6.2",
-    "uswds": "^1.6.3"
+    "react": "^15.5.4||^16.7.0",
+    "react-dom": "^15.6.2||^16.7.0",
+    "uswds": "^1.4.2"
   },
   "dependencies": {
     "@department-of-veterans-affairs/react-jsonschema-form": "^1.0.0",
--- a/test/js/widgets/SelectWidget.unit.spec.jsx
+++ b/test/js/widgets/SelectWidget.unit.spec.jsx
@@ -20,7 +20,7 @@ describe('Schemaform <SelectWidget>', () => {
         id="testing"
         onChange={onChange}
         options={{ enumOptions }}/>
-    );
+    ).dive(['SelectWidget']);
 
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -79,13 +79,13 @@
     "vets-json-schema": "git+https://github.com/department-of-veterans-affairs/vets-json-schema.git"
   },
   "peerDependencies": {
-    "react": "^15.5.4",
-    "react-dom": "^15.6.2",
-    "uswds": "^1.6.3"
+    "react": "^15.5.4||^16.7.0",
+    "react-dom": "^15.6.2||^16.7.0",
+    "uswds": "^1.4.2"
```
jbalboni: We've been on 1.4 on VA.gov without any issues, so I threw this change in here as well, because we've been seeing that warning for a while.

dmethvin-gov: Which warning is this? 1.6.9 fixed a vuln and that version range should use it.

test/js/widgets/SelectWidget.unit.spec.jsx
```diff
@@ -20,7 +20,7 @@ describe('Schemaform <SelectWidget>', () => {
         id="testing"
         onChange={onChange}
         options={{ enumOptions }}/>
-    );
+    ).dive(['SelectWidget']);
```
dmethvin-gov: Okay, so this is needed because `shallow` only renders one level and [SelectWidget exports a wrapped component](https://github.com/usds/us-forms-system/blob/a963c57f1ebcb39ba6c4c05d5ad6909d676f2198/src/js/widgets/SelectWidget.jsx#L58-L62) unlike the other widgets. So this goes down a level and renders the actual raw widget.

package.json
jbalboni: Sorry, an incorrect peer dependency warning in our repo. Do you what vulnerability 1.6.9 fixed over 1.4? Our security check isn't flagging any medium severity or higher issues.

test/js/widgets/SelectWidget.unit.spec.jsx
jbalboni: Yeah, I'm just not sure why the recompose change is forcing us to go a level down.

package.json
dmethvin-gov: https://github.com/uswds/uswds/releases/tag/v1.6.9

Is the problem that you're stuck on 1.4.2 and can't go above that for some reason? If so we can drop this down, but it would be better to use the latest if possible.

test/js/widgets/SelectWidget.unit.spec.jsx
dmethvin-gov: It doesn't look like it unwraps automatically anymore since https://github.com/acdlite/recompose/pull/473 ? And since their versioning hasn't gotten to 1.0.0 they don't have to worry about breaking changes in minor/patch releases. Yay.

package.json
jbalboni: We're currently trying to figure out how we should use USWDS in relation to the VA design system, and the last time someone checked there were some non-trivial changes we had to make to update from 1.4. So we're holding off for the moment. The warning is not a significant thing on our side, I just updated it because I was already changing peer deps. I'm fine changing it back if you all prefer it.

dmethvin-gov: USFS doesn't really have any dependency on USWDS that we test, we just put in the names it wants so dropping back is fine with me. You're pulling in the `.scss` files anyway rather than our built `.css` files, when I looked at the problem a few weeks ago it seemed like most library users would need to do that.

PR #329, opened 2019-01-07
add Required to ErrorableCheckbox

accbjt:
[SCREENREADER]: I have read the opt out policy checkbox needs "Required" added to label.

On Screen 2 of 2, the checkbox for I have read and accept the privacy policy * does not include a description of the asterisk or call it out as a required field like all the other fields. This is a content/text change, with screenshot attached below.

https://github.com/department-of-veterans-affairs/vets.gov-team/issues/13828

## Types of changes

## Checklist:


src/js/components/ErrorableCheckbox.jsx
```diff
@@ -33,7 +33,7 @@ class ErrorableCheckbox extends React.Component {
     // Calculate required.
     let requiredSpan = undefined;
     if (this.props.required) {
-      requiredSpan = <span className="form-required-span">*</span>;
+      requiredSpan = <span className="form-required-span">(*Required)</span>;
```
jbalboni: I think you could update `CheckboxWidget.jsx` as well. You'll also need to run `npm run build` and commit the changes in the lib folder.

accbjt: @jbalboni Okay

src/js/widgets/CheckboxWidget.jsx
```diff
@@ -11,7 +11,7 @@ export default function CheckboxWidget({
   onChange,
   options
 }) {
-  const requiredSpan = required ? <span className="form-required-span">*</span> : null;
+  const requiredSpan = required ? <span className="form-required-span">(*Required)</span> : null;
   const widgetClasses = classNames('form-checkbox', options.widgetClassNames);
```
accbjt: @jbalboni I updated it.

dmethvin-gov:
As discussed on Slack, I think we're fine with incorporating this into 1.x so that the VA will have consistency now. To be clear, though, it's likely to be a dead end.

Longer term it is highly likely that if USFS is opinionated on this issue it will lean towards the USWDS 2.0 recommendation to mark *optional* fields.  I would think that required fields are more common than optional ones in most forms.  Denoting just the optional fields reduces both visual and audio clutter.

I also doubt this is a place where we'd just "add a flag to do it either way" because it would be really confusing to have government websites varying on something this fundamental.

jbalboni:
@dmethvin-gov Yep, that makes sense. I would expect this library to always follow USWDS and if the VA design system diverges from that then we have to figure what the right approach to USFS is on the VA.gov side.

PR #334, opened 2019-01-15
Fix typo in file field

cvalarida:
I'm writing an e2e tester that enters data into forms from a json file by using the field `id`s or `name`s to determine the data path in the json data. I discovered this little typo as a result of that. My tester couldn't find the data because `'atachmentId' !== 'attachmentId'`.

## Types of changes

## Checklist:


lib/js/fields/ArrayField.js
```diff
@@ -70,8 +70,8 @@ var ArrayField = function (_React$Component) {
      */
 
     _this.state = {
-      editing: props.formData ? props.formData.map(function () {
-        return false;
+      editing: props.formData ? props.formData.map(function (item, index) {
+        return !(0, _validation.errorSchemaIsValid)(props.errorSchema[index]);
```
cvalarida: :thinking: Looks like this is from [my last PR](https://github.com/usds/us-forms-system/pull/321/files). Not sure why it's showing up as a new diff. :confused:

src/js/fields/FileField.jsx
```diff
@@ -122,10 +122,10 @@ export default class FileField extends React.Component {
               });
               const itemSchema = schema.items[index];
               const attachmentIdSchema = {
-                $id: `${idSchema.$id}_${index}_atachmentId`
+                $id: `${idSchema.$id}_${index}_attachmentId`
               };
               const attachmentNameSchema = {
-                $id: `${idSchema.$id}_${index}_atachmentName`
+                $id: `${idSchema.$id}_${index}_attachmentName`
```
cvalarida: These two things are what I was _actually_ changing here.

lib/js/fields/ArrayField.js
dmethvin-gov: Probably the PR was based off master before that PR landed?

dmethvin-gov: Oh, it's because only the /src change made it in and not the /lib one. 😢  Can you run the build and commit the FileField.js file that it updates?

src/js/fields/FileField.jsx
dmethvin-gov: This has been misspelled for a while so changing the name could be considered a breaking change. I don't think that's a problem in reality though.

lib/js/fields/ArrayField.js
dmethvin-gov: Oh nevermind, it looks like the correct FileField change is there already.

cvalarida:
@dmethvin-gov Hey! Thanks for taking a look. Is there something else I should do to get this merged in? I definitely don't want to cause a breaking change for anybody else who might be using the stable branch, but we (VA.gov) also aren't ready to use versions from `master` yet.

cvalarida:
Thanks, @dmethvin-gov! :raised_hands:

PR #338, opened 2019-01-18
Fix expansion of sample SS-5 form

### LINKED ISSUE #337: Sample SS-5 form - problem with question expansion
I'm running the example form here: https://github.com/usds/us-forms-system/blob/28b9c581a6d48ca8656bc5481c34f17ffc439585/docs/examples/SS-5--form.js

When I select "No" or "Do Not Know" for the question "Has the applicant previously filed for a Social Security card?", the dropdown information "Enter information from previous Social Security card" shows up.

image

**Expected behavior**: Only when I select "Yes" for that question should the dropdown information "Enter information from previous Social Security card" show up.

epicfaace:
Fixes #337 -- expands view:previouslyFiled only when previouslyFiledForSSN is equal to "Yes".

### DIFF
--- a/docs/examples/SS-5--form.js
+++ b/docs/examples/SS-5--form.js
@@ -129,7 +129,8 @@ const formConfig = {
             },
             'view:previouslyFiled': {
               'ui:options': {
-                expandUnder: 'previouslyFiledForSSN'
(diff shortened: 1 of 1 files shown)


dmethvin-gov:
Thanks!

PR #340, opened 2019-02-06
Send cookies with form submission request

erikphansen:
We need to send cookies with the form submission request (and stop using the user token).

Addresses department-of-veterans-affairs/vets.gov-team#16752 where completed/submitted forms were showing as being "in progress"

## Types of changes

## Checklist:


test/js/actions.unit.spec.js
```diff
@@ -129,31 +129,6 @@ describe('Schemaform actions:', () => {
 
       return promise;
     });
-    it('should submit with auth header', () => {
```
erikphansen: This test is no longer needed. _BUT_ it would be ideal to add a test that either:

- checks that Cookies are added to the request, or...
- checks that we configured the XHR with `withCredentials`

src/js/actions.js
```diff
@@ -124,11 +123,7 @@ function submitToUrl(body, submitUrl, trackingPrefix) {
 
     req.setRequestHeader('X-Key-Inflection', 'camel');
     req.setRequestHeader('Content-Type', 'application/json');
-
```
dmethvin-gov: This is most likely just a leftover from the VA-to-USFS migration since USFS never had any concept of users.

test/js/actions.unit.spec.js
dmethvin-gov: Removing it is fine since it really shouldn't have been there anyway, I did it in a [later commit](https://github.com/usds/us-forms-system/commit/37c569781d188be421eed28e9b5987aeea3e6370#diff-731cfe914af9eff1a334a7c9ab16bcd6L132) for 2.0.  There's not an easy way to verify that cookies were sent to the server so let's punt on that.

src/js/actions.js
```diff
@@ -124,11 +123,7 @@ function submitToUrl(body, submitUrl, trackingPrefix) {
...
-    const userToken = _.get('sessionStorage.userToken', window);
-    if (userToken) {
-      req.setRequestHeader('Authorization', `Token token=${userToken}`);
-    }
+    req.withCredentials = true;
```
dmethvin-gov: Just a note that this works on IE10 and later. I see markup on the VA site that seems to be trying to support IE9 and below but as far as I can tell it can't possibly work there. Not sure if there is an official browser support goal but IE11 is fine with me.

PR #341, opened 2019-02-08
update README

fatima-gov:
add note that project is no longer active.


fatima-gov:
@dmethvin-gov I think that's great!

dmethvin-gov:
Closed in favor of #343

PR #342, opened 2019-02-09
Allow for non-form-page routes in a form system app

erikphansen:
I have run into a situation where I want to add an arbitrary page/route
to the Health Care Application. It makes sense for this new page to
actually live in the HCA form system React app, but there currently
isn't a good way to make that work.

It's relatively straightfoward to add a new page/route to the list of
routes generated by createRoutesWithSaveInProgress() but that page will
be rendered with the title and form progress bars that get added to
every page in the form. This change makes is possible to add a list of
paths in the form config (under a new `additionalRoutes` property) and
the FormApp component won't FormTitle, FormNav, or extra wrapper markup
around the child component.

## Types of changes

## Checklist:


erikphansen:
You can see an example of an app using this new config here:
- The changes to the form config JS, adding `additionalRoutes`: https://github.com/department-of-veterans-affairs/vets-website/blob/cea659fa80a139eb9e0aafe263839d90e0c9f056/src/applications/hca/config/form.js#L169-L175
- The changes to the app's routes JS, making use of the `additionalRoutes` added to the form config: https://github.com/department-of-veterans-affairs/vets-website/blob/cea659fa80a139eb9e0aafe263839d90e0c9f056/src/applications/hca/routes.jsx

dmethvin-gov:
It looks like the use case is to add more introduction pages to the front of the form. It almost has to be that way since it would be hard to insert non-form pages into the middle of the form, technically the order of the form pages isn't something the code should be prying into at the `FormApp` level. Also, adding non-navigable pages in the middle of the form is something that shouldn't be done without exploring the UX issues.

There's already a feature (**undocumented!**) to allow multiple intro pages, via the `endsWith("introduction")` check. The setup insists in naming the first page "introduction" but you could follow that with "id-introduction" and so forth. Could that be used? I'm assuming someone put in the `endsWith` to allow this kind of thing.

erikphansen:
Thanks for the response, Dave. I had stumbled on the fact that simply adding `introduction` to the end of the path name does the trick. The route I need to add to the HCA is sort of an introduction page. It's also entirely optional; the user will only be shown the route if they aren't logged in (and possibly another scenario that depends on the LOA).

FWIW, this is the code that adds the additional pages to the app: https://github.com/department-of-veterans-affairs/vets-website/blob/cea659fa80a139eb9e0aafe263839d90e0c9f056/src/applications/hca/routes.jsx#L24

dmethvin-gov:
If `X-introduction` isn't a good name for the page, perhaps we could add an `endsWith("instructions")` to the list of special cases? That should work unless some existing form has a form page ending with that word. I'm not a big fan of the whole "ends with" approach here but the pattern is already established and no doubt can't be changed without breaking something already deployed. Adding another word to that list seems better than creating a whole new mechanism that overlaps it.

erikphansen:
`id-introduction` will work for the path name. Part of the reason I didn't plow ahead with that was that it felt hacky and not as future-proof, basically because I also didn't like the `endsWith` approach :)

Let's drop this PR and I'll move forward with using a path name that ends in `introduction` for my additional route.

PR #343, opened 2019-02-12
Update README.md

dmethvin-gov:
Update status but let people know they can reach us if they're interested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,9 @@
 
 The US Forms System is an open source library and set of guiding principles for building complex web-based forms using [React](https://reactjs.org), the [JSON Schema standard](http://json-schema.org/), and the [US Web Design System](https://designsystem.digital.gov/).
 
-***THIS IS A WORK IN PROGRESS.***
+### This project is not being actively maintained.
+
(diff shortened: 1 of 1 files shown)


dmethvin-gov:
@easherma we haven't really worked out a plan. If someone wants to do a significant amount of refactoring on the existing code I don't think this repo would be the best place to host it, since we would have limited time to participate and don't want to be blocking the pace of progress others might want to take.

The VA team created the code out of a need they had, then we migrated it out of their repo into its own library. The current code is still tied in many ways to their view of forms and largely meets those needs. As a proof of concept for a generalized library, I think it has been extremely valuable.

The conclusion we reached from this exercise is that right now the library imposes design choices that don't appeal to a wide enough set of potential users. We can try to modify the existing code base to address some of those, but it would be better to step back and think about what our goals are for an independent library, who the target audience is, and what level of flexibility we need. 

At some point soon, the VA team plans to fork the changes being made in the 1.x-stable branch back into their own repo to continue adding changes and fixes. I expect those changes will remain focused on their needs.

We're planning to write up these thoughts in more detail during the next month. After that we can reference the documents here and get your thoughts about where a reimagined library might go. There is no doubt that everyone (not just the US government) needs a better way to create forms!

PR #351, opened 2019-03-12
Update dependencies to remove vulns

dmethvin-gov:
extend: CVE-2018-1649
just-extend: CVE-2018-16489


dmethvin-gov:
I turned off PR approvals so I could land this ASAP. Since we're most likely just doing security vuln changes in the near term it would be convenient to keep them turned off for now.

